PR #25, opened 2018-02-27
Added support for variable exposure rate and HRSC

### LINKED ISSUE #23: Add support for variable line scan rates
Many cameras such as HRSC have variable line scan rates. The current line scan camera model is setup for constant line scan rates. In order to support HRSC and other variable rate cameras, we need to add the ability to compute the image time from a table/list/vector/etc. of line scan rates.

jessemapel:
These changes are not backwards compatible with existing LS camera states because of how the exposure/integration time is stored.

This also removes the functional component of the "line flipping" code, because it was impacting HRSC line time calculations.

resolves #23


include/genericls/UsgsAstroLsStateData.h
```diff
@@ -85,7 +85,9 @@ class UsgsAstroLsStateData
    int          m_PlatformFlag;                   // 7
    int          m_AberrFlag;                      // 8
    int          m_AtmRefFlag;                     // 9
-   double       m_IntTime;                        // 10
+   std::vector<double> m_IntTimeLines;
+   std::vector<double> m_IntTimeStartTimes;
+   std::vector<double> m_IntTimes;
```
jessemapel: These could be changed to a vector of structs. This would make some things a bit safer and better match ISIS.

src/UsgsAstroLsPlugin.cpp
```diff
@@ -421,9 +423,22 @@ std::string UsgsAstroLsPlugin::convertISDToModelState(
    state.m_PlatformFlag = atoi(image_support_data.param("PLATFORM").c_str());
    state.m_AberrFlag = atoi(image_support_data.param("ABERR").c_str());
    state.m_AtmRefFlag = atoi(image_support_data.param("ATMREF").c_str());
-   state.m_IntTime = atof(image_support_data.param("INT_TIME").c_str());
    state.m_StartingEphemerisTime = atof(image_support_data.param("STARTING_EPHEMERIS_TIME").c_str());
    state.m_CenterEphemerisTime = atof(image_support_data.param("CENTER_EPHEMERIS_TIME").c_str());
+   if (image_support_data.param("NUMBER_OF_INT_TIMES").empty()) {
```
jessemapel: This if could be changed to check if INT_TIME has more than one value, instead of looking for the NUMBER_OF_INT_TIMES parameter. It could also be changed to look for any value greater than 1.

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -108,7 +109,9 @@ void UsgsAstroLsSensorModel::updateState()
    _data.m_HalfSwath = _data.m_Gsd * _data.m_TotalSamples / 2.0;
 
    // Compute half time duration
-   _data.m_HalfTime = _data.m_IntTime * _data.m_TotalLines / 2.0;
+   double fullImageTime = _data.m_IntTimeStartTimes.back() - _data.m_IntTimeStartTimes.front()
+                          + _data.m_IntTimes.back() * (_data.m_TotalLines - _data.m_IntTimeLines.back());
+   _data.m_HalfTime = fullImageTime / 2.0;
```
jessemapel: Depending on what the half time actually is, this could be wrong. I assumed half time is the center time in the image because the time at the center pixel is covered by center ephemeris time. For a fixed rate line scan camera, those two values are identical.

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -870,22 +873,27 @@ double UsgsAstroLsSensorModel::getImageTime(
 {
    // Flip image taken backwards
    double line1 = image_pt.line;
-   if (_data.m_ImageFlipFlag == 1)
-      line1 = _data.m_TotalLines - 1 - image_pt.line;
```
jessemapel: As stated in the PR description, I removed the functional part of the image flip flag, but left the computation code in. Depeneding on how we want to handle removing that it can be added to this pr.

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -870,22 +873,27 @@ double UsgsAstroLsSensorModel::getImageTime(
...
-      + _data.m_StartingEphemerisTime;
+   // These calculation assumes that the values in the integration time
+   // vectors are in terms of ISIS' pixels
+   auto referenceLineIt = std::upper_bound(_data.m_IntTimeLines.begin(),
+                                           _data.m_IntTimeLines.end(),
+                                           lineUSGSFull);
+   if (referenceLineIt != _data.m_IntTimeLines.begin()) {
+      --referenceLineIt;
+   }
+   size_t referenceIndex = std::distance(_data.m_IntTimeLines.begin(), referenceLineIt);
```
jessemapel: If the exposure time table was a vector of structs/arrays instead of 3 vectors, we wouldn't need to convert the iterator to an index, we could just use the iterator directly for the time calculation.

src/UsgsAstroLsStateData.cpp
```diff
@@ -345,7 +366,9 @@ void UsgsAstroLsStateData::setState(const std::string &stateString )
    m_PlatformFlag = j["STA_PLATFORM_FLAG"];
    m_AberrFlag = j["STA_ABERR_FLAG"];
    m_AtmRefFlag = j["STA_ATMREF_FLAG"];
-   m_IntTime = j["STA_INT_TIME"];
+   m_IntTimeLines = j["STA_INT_TIME_LINES"].get<std::vector<double>>();
+   m_IntTimeStartTimes = j["STA_INT_TIME_START_TIMES"].get<std::vector<double>>();
+   m_IntTimes = j["STA_INT_TIMES"].get<std::vector<double>>();
```
jessemapel: I'm not sure how the JSON library will handle serializing a vector of structs/arrays.  This is the main reason I started with 3 vectors.

src/UsgsAstroLsPlugin.cpp
jlaura: Since we are carrying a bunch of `NUMBER_OF_*` inside the ISD this is a 'canonical' implementation.  I would keep this way.

src/UsgsAstroLsPlugin.cpp
```diff
@@ -421,9 +423,22 @@ std::string UsgsAstroLsPlugin::convertISDToModelState(
...
+   }
+   else {
+     int numIntTimes = atoi(image_support_data.param("NUMBER_OF_INT_TIMES").c_str());
+     for (int i = 0; i < numIntTimes; i++) {
+       state.m_IntTimeLines.push_back(atof(image_support_data.param("INT_TIME", i*3).c_str()));
+       state.m_IntTimeStartTimes.push_back(atof(image_support_data.param("INT_TIME", i*3 + 1).c_str()));
+       state.m_IntTimes.push_back(atof(image_support_data.param("INT_TIME", i*3 + 2).c_str()));
+     }
+   }
+   state.m_CenterEphemerisTime = atof(image_support_data.param("CENTER_EPHEMERIS_TIME").c_str());
```
jlaura: this whole block demonstrates where a more friendly interface to the ISD would be nice.  No 'clean' way to implement for now.

src/UsgsAstroLsStateData.cpp
jlaura: https://github.com/nlohmann/json#arbitrary-types-conversions

Looks pretty straight forward?

src/UsgsAstroLsSensorModel.cpp
jlaura: I also don't know.  Since we have a working HRSC, I assume that this is correct?

src/UsgsAstroLsSensorModel.cpp
jlaura: :+1: go with a vector of structs.

src/UsgsAstroLsStateData.cpp
```diff
@@ -147,7 +149,9 @@ std::string UsgsAstroLsStateData::toJson() const {
         {STATE_KEYWORD[STA_PLATFORM_FLAG], m_PlatformFlag},
         {STATE_KEYWORD[STA_ABERR_FLAG], m_AberrFlag},
         {STATE_KEYWORD[STA_ATMREF_FLAG], m_AtmRefFlag},
-        {STATE_KEYWORD[STA_INT_TIME], m_IntTime},
+        {STATE_KEYWORD[STA_INT_TIME_LINES], m_IntTimeLines},
```
jlaura: This then will, I think, need to be altered to use an vector of arbitrary objects.

src/UsgsAstroLsStateData.cpp
```diff
@@ -214,7 +218,24 @@ std::string UsgsAstroLsStateData::toString() const
    state_stream << STATE_KEYWORD[STA_PLATFORM_FLAG]     << " " << m_PlatformFlag      << "\n";  // 9
    state_stream << STATE_KEYWORD[STA_ABERR_FLAG] << " " << m_AberrFlag         << "\n";  // 10
    state_stream << STATE_KEYWORD[STA_ATMREF_FLAG]<< " " << m_AtmRefFlag        << "\n";  // 11
-   state_stream << STATE_KEYWORD[STA_INT_TIME]   << " " << m_IntTime           << "\n";  // 12
+   state_stream << STATE_KEYWORD[STA_INT_TIME_LINES]    << " " ;
```
jlaura: In terms of maintainability, I think we need to decide if serialization to JSON is sufficient and we can avoid also maintaining serialization `toString()`?

src/UsgsAstroLsStateData.cpp
jlaura: link below.

tests/conftest.py
```diff
@@ -41,6 +41,27 @@ def ctx_model():
...
+    plugin = cam.genericls.Plugin()
+    return plugin.from_isd(csm_isd, plugin.modelname(1))
+
+@pytest.fixture
+def hrsc_stereo_2_model():
+    path = os.path.join(data_path, 'h0232_0000_s22_keywords.lis')
+    csm_isd = isd.Isd.read_socet_file(path)
+    plugin = cam.genericls.Plugin()
+    return plugin.from_isd(csm_isd, plugin.modelname(1))
+
```
jlaura: are you thinking that these fixtures will be widely used or are they tightly coupled to the classes with like names?  If the latter, perhaps move them inside the class since no other test is likely to need these fixtures?

jlaura:
@jessemapel Major comment is just, use a vector of structs so that the implicit relationship is explicit. I am fine merging this an opening an issue to refactor to a vector of structs or updating this PR since the code is fully functional.  Whatever you want to do on your side.

jessemapel:
@jlaura Go ahead and merge this in. I tried to do some work with the struct but am having a bunch of issues with the JSON library. At this point, I think this should just be merged in and that should be an open issue.

PR #29, opened 2018-03-05
Fixed look vectors with fractional lines

### LINKED ISSUE #26: Generic line scan camera does not correctly calculate sub-pixel look vectors
The `losToEcf` method in the generic line scan camera model computes a look vector from an image coordinate.  It correctly accounts for fractional samples, but it doesn't account for fractional lines. Regardless of what the entered line value is, it is always assumed to be the center of the nearest line.

This method is used in both image to ground and ground to image.

jessemapel:
The fix for look vector calculations is fairly simple. It did reveal a problem with ground to image iteration being slow. Before, it only had to converge within a pixel.  Now we are requiring far higher accuracy and it is taking ~15 iterations.

Fixes #26


src/UsgsAstroLsSensorModel.cpp
```diff
@@ -317,8 +317,8 @@ csm::ImageCoord UsgsAstroLsSensorModel::groundToImage(
             "Divide by zero.",
             "UsgsAstroLsSensorModel::groundToImage");
       }
-      lineTemp += (dLine / det / DELTA_IMAGE);
-      sampleTemp += (dSamp / det / DELTA_IMAGE);
+      lineTemp += (dLine / det * DELTA_IMAGE);
+      sampleTemp += (dSamp / det * DELTA_IMAGE);
```
jessemapel: This was a math error that was present in the original code. DELTA_IMAGE was just always 1, so it didn't matter.

tests/test_genericls.py
```diff
@@ -82,8 +82,8 @@ def test_ground_to_image(self, hrsc_nadir_model, image, ground):
         y, x = hrsc_nadir_model.groundToImage(*ground)
         iy, ix, _ = image
 
-        assert x == pytest.approx(ix, abs=0.5)
-        assert y == pytest.approx(iy, abs=0.5)
+        assert x == pytest.approx(ix, abs=1)
+        assert y == pytest.approx(iy, abs=1)
```
jessemapel: These tests had to be loosened up because the ground to image precision is now much higher and the actual results are different.

jessemapel: We should change these from comparing against ISIS to comparing against CSM.

jessemapel:
See #30 for more information about ground to image iteration in the line scan camera.

PR #31, opened 2018-03-07
ISD keyword update

twilson271828:
This is the branch Kristin/myself were working on Friday.  She has updates to this on her laptop, but is having difficulties in finding a reliable wifi connection, so I told her I would make the PR and when she gets that sorted out she will send in updates.


jessemapel:
What is the isd_notes.json file?

schemas/isd.json
```diff
@@ -0,0 +1,242 @@
...
+               "type": "array",
+               "items" : {
+                  "type": "number",
+                  "minItems": 4 ,
+                  "maxItems": 4
+           }
+          }
+      },
+
+       "DETECTOR_LINE_RATE": {
```
jessemapel: I'm not sure if this is the correct name for this. Previously this was Integration Time. ISIS calls this Line Rate. It's also been called Exposure Rate and Line Scan Rate. I lean towards Line Scan Rate, but I'm not sure how well that matches image labels.

schemas/isd.json
```diff
@@ -0,0 +1,242 @@
...
+	                     },
+	                     "EXPOSURE_TIME": {
+		                       "type": "number"
+	                     },
+	          "required": [ "EXPOSURE_TIME"]
+	                 }
+	              }
+      },
+       "DETECTOR_SAMPLE_SUMMING": {
+           "description": "The summing in the sample-direction",
```
jessemapel: We should specify that summing of 1 is no summing, summing of 2 is 2 pixel summing, 4 is 4 pixel summing, etc. Not all missions use this format.  For some missions, they use 0 is no summing, 1 is 2 pixel, 2 is 4 pixel, etc.

schemas/isd.json
```diff
@@ -0,0 +1,242 @@
...
+              "maxItems": 2
+           }
+       },
+      "INTERPOLATION_METHOD": {
+           "description": "The type of interpolation method to use.",
+           "type": "string"
+       },
+
+
+   "required": ["FOCAL_LENGTH", "BORESIGHT","CCD_CENTER","IMAGE_LINES","IMAGE_SAMPLES" ]
```
jessemapel: This is not all of the required parameters. We have to have ephemeri and the focal plane to pixel conversions at least.  There's probably more I'm missing.

schemas/isd.json
```diff
@@ -0,0 +1,242 @@
...
+           "description": "The line on the detector where the data starts",
+           "type": "number"
+       },
+       "FOCAL2PIXEL_X": {
+           "description": "Transformation x coefficients from focal plane (mm) to pixel coordinates",
+           "type": "array",
+	          "items": {
+                 "type": "number"
+            }
+       },
```
jessemapel: This is a 3 element array so it should be included in the schema

schemas/isd.json
```diff
@@ -0,0 +1,242 @@
...
+       "FOCAL2PIXEL_Y": {
+           "description": "Transformation y coefficients from focal plane (mm) to pixel coordinates",
+           "type": "array",
+	          "items": {
+                 "type": "number"
+            }
```
jessemapel: This is also a 3 element array

schemas/isd.json
```diff
@@ -0,0 +1,242 @@
...
+       "STARTING_DETECTOR_SAMPLE": {
+           "description": "The sample on the detector where the data starts",
+           "type": "number"
+       },
+       "STARTING_DETECTOR_LINE": {
+           "description": "The line on the detector where the data starts",
+           "type": "number"
+       },
+       "FOCAL2PIXEL_X": {
+           "description": "Transformation x coefficients from focal plane (mm) to pixel coordinates",
```
jessemapel: This converts to detector pixels, it should be mentioned in the description because it is not clear if it converts from focal to detector pixels or focal to image pixels.

schemas/isd.json
```diff
@@ -0,0 +1,242 @@
...
+           "type": "array",
+	          "items": {
+                 "type": "number"
+            }
+       },
+       "FOCAL2PIXEL_Y": {
+           "description": "Transformation y coefficients from focal plane (mm) to pixel coordinates",
```
jessemapel: Same comments as FOCAL2PIXEL_X. This should specify detector pixels.

kberryUSGS:
@jessemapel Thanks for all the comments. The `isd_notes.json` should not be in this PR. I must have accidentally added _everything_ with local changes instead of just what I intended. The only file that should be in this PR is `isd.json`.

kberryUSGS:
But to answer your actual question `isd_notes.json` is the version of the `json` file I was taking notes in during a meeting in which people were giving feedback about the proposed schema.

schemas/isd.json
```diff
@@ -0,0 +1,242 @@
...
+           "type": "array",
+	    "items": {
+                 "type": "number",
+		 "description": "Component of the boresight vector",
+		 "minItems": 3,
+                 "maxItems": 3
+            }
+       },
+       "DETECTOR_CENTER": {
+           "description": "Center of the detector in pixel coordinates",
```
jessemapel: What is the order of the pixel coordinates? Line, Sample?

jessemapel:
In general, why are some keywords all caps and others all lowercase?

schemas/isd.json
```diff
@@ -0,0 +1,242 @@
...
+       },
+       "FOCAL_LENGTH_MODEL": {
+            "description":  "A key to get information about time/temp dependent focal length models",
+            "type": "object",
+            "properties": {
+                "key": {"type": "string"},
+                "focal_length": {"type": "number"},
+                "focal_epsilon":{"type": "number"}
+            },
+   	"required": ["focal_length"]
```
jessemapel: Should "key" be required also if we're going to use it for parsing?  We could default it to constant if we don't want to require it.

schemas/isd.json
```diff
@@ -0,0 +1,242 @@
...
+       },
+       "IFOV": {
+           "description": "Instantaneous field of view in degrees",
+           "type": "number"
+       },
```
jessemapel: I don't think we actually need this. The IFOV is usually not used by the camera models.

schemas/isd.json
```diff
@@ -0,0 +1,242 @@
...
+
+       "RADII": {
+           "description": "The radii of the target.",
+           "type": "array",
+	    "items": {
+                 "type": "number",
+		 "minItems": 1,
+                 "maxItems": 3
```
jessemapel: CSM currently only supports 2, so we should probably cap that here.

schemas/isd.json
```diff
@@ -0,0 +1,242 @@
...
+            }
+       },
```
jessemapel: The line scan camera currently takes semi major radius and eccentricity.  Do we want to support that format also?

schemas/isd.json
kberryUSGS: :+1:

schemas/isd.json
kberryUSGS: You're right this is definitely not all the required parameters! This was something that needed to be updated, thank you.

schemas/isd.json
kberryUSGS: :+1:

schemas/isd.json
kberryUSGS: :+1:

schemas/isd.json
kberryUSGS: :+1:

schemas/isd.json
kberryUSGS: :+1:

schemas/isd.json
kberryUSGS: Good point. I remember discussion about supporting triaxial ellipsoids, but did not remember the conclusion / where this discussion left off.

schemas/isd.json
kberryUSGS: Someone else correct me if I'm wrong but -- if I remember correctly, this came up during an initial review of this schema during a standup and we decided to not directly support the semi-major radius + eccentricity format. I'm not sure where the responsibility to deal with this lies -- the ISD generator ? The camera model itself? (Which would then require it to be updated.)

schemas/isd.json
kberryUSGS: :+1:

schemas/isd.json
kberryUSGS: I'm not sure what the best name for this is either. `LINE_SCAN_RATE` sounds good to me.

schemas/isd.json
```diff
@@ -0,0 +1,242 @@
...
+       "BORESIGHT": {
+	   "description": "Boresight (x,y,z) vector",
+           "type": "array",
+	    "items": {
+                 "type": "number",
+		 "description": "Component of the boresight vector",
+		 "minItems": 3,
+                 "maxItems": 3
+            }
+       },
```
jessemapel: What is this for? The detector center and focal2pixel parameters cover all of the math needed to account for the boresight.

schemas/isd.json
jessemapel: It sounds like converting to semi_major semi_minor will be the responsibility of the ISD generator.

jessemapel:
All of the places that have a "keys" element we should document the valid options for that key. Is there some functionality in the validation to check that?

schemas/isd.json
kberryUSGS: I was going to initially answer that it was _used_ by the `UsgsAstroFrameSensorModel`, but you're right, it's not. It's _required_ by it, but not actually used by it, so this looks like it can be eliminated.

schemas/isd.json
```diff
@@ -0,0 +1,242 @@
...
+
+       "STARTING_DETECTOR_SAMPLE": {
+           "description": "The sample on the detector where the data starts",
+           "type": "number"
+       },
+       "STARTING_DETECTOR_LINE": {
+           "description": "The line on the detector where the data starts",
+           "type": "number"
+       },
+       "FOCAL2PIXEL_X": {
```
jessemapel: These should be named FOCAL2PIXEL_SAMPLES because this converts x and y to samples

schemas/isd.json
```diff
@@ -0,0 +1,242 @@
...
+           "description": "Transformation x coefficients from focal plane (mm) to pixel coordinates",
+           "type": "array",
+	          "items": {
+                 "type": "number"
+            }
+       },
+       "FOCAL2PIXEL_Y": {
```
jessemapel: Same as previous, this should be called FOCAL2PIXEL_LINES because it converts x and y to lines

schemas/isd.json
jessemapel: That is based on the line scan. I'm not sure about the framer?

schemas/isd.json
jessemapel: Also, we should note the coefficient order is constant, x , y.

schemas/isd.json
jessemapel: If this is undecided, we should go with line, sample because that's the CSM standard.

schemas/isd.json
```diff
@@ -0,0 +1,242 @@
...
+               "type": "array",
+	              "items" : {
+                    "type": "number",
+                    "minItems": 3,
+                    "maxItems": 3
+		                }
+	                 }
+	     },
+       "SUN_POSITION" : {
+           "description": "(v_x, v_y, v_z) velocity of sun in body-fixed frame.",
```
jessemapel: This description is a duplicate of the velocity description.

schemas/isd.json
```diff
@@ -0,0 +1,242 @@
...
+	              "items" : {
+                  "type": "number",
+                  "minItems": 3,
+                  "maxItems": 3
+		        }
+	         }
+	     },
+
+       "ORIENTATION" : {
+           "description": "Orientation of the sensor provided in quaternions",
```
jessemapel: We should note the quaternion format. The "scalar" quaternion is the first, then the three "rotation" quaternions. This is different from what NAIF returns with SPICE. The line scan model uses the Alternate Style described in this NAIF white paper ftp://naif.jpl.nasa.gov/pub/naif/misc/Quaternion_White_Paper/Quaternions_White_Paper.pdf.

schemas/isd.json
```diff
@@ -0,0 +1,242 @@
...
+                      "LINE_START_TIME": {
+ 	                        "type": "number"
+	                     },
+	                     "EXPOSURE_TIME": {
+		                       "type": "number"
+	                     },
+	          "required": [ "EXPOSURE_TIME"]
+	                 }
+	              }
+      },
```
jessemapel: This table could have thousands of entries in it. I'm concerned about having to have `{START_LINE: value, LINE_START_TIME: value, EXPOSURE_TIME:}` duplicated for each entry. Would it be reasonable to convert this to a 2d array like the ephemeri?

schemas/isd.json
jessemapel: It's also worth noting that exposure time and integration time are generally the same thing, but scan rate could be different.  For some line scan cameras, there's a small but relevant window of time in-between the end of one line and the start of the next.

jessemapel: The values in this table are kind of used as both.

schemas/isd.json
```diff
@@ -0,0 +1,242 @@
...
+       "NUMBER_OF_QUATERNIONS": {
+           "description": "The number of quaternions",
+           "type": "number"
+       },
+       "DT_EPHEM": {
+           "description": "The time between each ephemeris point.",
+           "type": "number"
+       },
+       "T0_EPHEM": {
+           "description": "First ephemeris time - center ephemeris time",
```
jessemapel: I think this should be first quaternion ephemeris time - center image time

schemas/isd.json
```diff
@@ -0,0 +1,242 @@
...
+           "type": "number"
+       },
+       "DT_QUAT": {
+           "description": "Time spacing of quaternions.",
+           "type": "number"
+       },
+       "T0_QUAT": {
+           "description": "First quaternion time - center quaternion time.",
```
jessemapel: Same as ephemeri, I think this should be first quaternion time - center image time

schemas/isd.json
thareUSGS: Agreed capped at two for now in the v3.0.3 spec. Not that this will make everyone comforted but our current explanation: https://github.com/USGS-Astrogeology/CSM-CameraModel/wiki/CSM-and-triaxial

schemas/isd.json
thareUSGS: +1 for using semi_major, semi_minor. Eccentricity can be calculated.

schemas/isd.json
thareUSGS: +1

schemas/isd.json
thareUSGS: LINE_SCAN_RATE works for me. In all descriptions for "RATE"s can we list the unit of time (e.g. in seconds).

kberryUSGS: Yes, that's a good point. We'll add 'seconds' to the description.

schemas/isd.json
kberryUSGS: :+1: Let's do that, then.

schemas/isd.json
kberryUSGS: :+1:

schemas/isd.json
kberryUSGS: :+1:

schemas/isd.json
kberryUSGS: :+1:

schemas/isd.json
kberryUSGS: :+1:

schemas/isd.json
kberryUSGS: :+1:

schemas/isd.json
kberryUSGS: :+1:

schemas/isd.json
kberryUSGS: Yes, I will do that.

schemas/isd.json
kberryUSGS: :+1:

jessemapel:
It looks like the line scan plugin got reverted to the wrong commit. It has reverted the changes for variable line scan rates and HRSC.

jessemapel:
Is this schema intended to be for both the generic line scan and generic framing models? The only real difference between the two is how their exposure times and image times work.

kberryUSGS:
@jessemapel :+1: Thanks. I was wondering why this failed. 

It is intended to be for both linescanners and framing camera. It's part of why we had some kind of odd name in the past -- it was an attempt to cover both the linescanner and single exposure time keywords in one, but perhaps this isn't the correct approach. I agree that scan rate doesn't really make sense for a framer.

kberryUSGS:
@jessemapel (or anyone else?) can you think of a word that is covers both (exposure/integration time) and also line scan rate ? The word equivalent of a parent class for both of those concepts?

jessemapel:
@kberryUSGS I'd just got with integration time for right now. So, I guess ignore all of my previous comments about it! It's kind of a minor nit-pick for all but a couple cases.

kberryUSGS:
That works! The other cases _are_ definitely important, though, so we'll have to figure out how to deal with them at some point.

schemas/isd.json
```diff
@@ -0,0 +1,236 @@
...
+            "items" : {
+               "type": "number",
+               "minItems": 4 ,
+               "maxItems": 4
+            }
+         }
+      },
+      "line_scan_rate": {
+         "description": "The line rates of the detector map with the associated start line and time",
+         "type": "array",
```
jessemapel: To make this work for both the line scan and framing camera, it'd be great if this would take either an array of objects or a single value. That way the framing camera and constant rate line scanners can just pass a single value.

kberryUSGS: :+1:

schemas/isd.json
```diff
@@ -0,0 +1,236 @@
...
+         "items": {
+            "type": "array",
+             "items" : {
+               "type": "number",
+               "default: "-1",
```
jessemapel: This line needs a quote at the end of default

schemas/isd.json
```diff
@@ -0,0 +1,236 @@
...
+               "maxItems": 3
+            }
+         }
+      },
+      "sensor_orientation" : {
+         "description": "Orientation of the sensor provided in quaternions",
+         "type": "array",
+         "items": {
+            "type": "array",
+            "description": "Each quaternion is of the form: (scalar, vector, vector, vector)"
```
jessemapel: I'm not sure what is meant by vector in the context of a quaternion? Each quaternion is a vector of 4 scalars.  The important distinction is what the 4 values represent.  It may be better to link to an outside source that defines this for us because it takes some explanation. I'm going to try and find a good concise source.

kberryUSGS: The naif whitepaper you linked to earlier I think explained both quaternions and all of the different ways quaternion formats may be different from each other due to the lack of a standard specification pretty well.

kberryUSGS: @jessemapel Thanks for pointing that out: it should be 'vector component', rather than 'vector'. IIRC, that whitepaper referred to the components of a quaternion as (scalar, vector component, vector component, vector component).

schemas/isd.json
```diff
@@ -0,0 +1,236 @@
...
+       },
+       "starting_ephemeris_time": {
+           "description": "The starting ephemeris time of the image",
+           "type": "number"
+       },
+       "focal_length_model": {
+           "description":  "The static focal length or the information needed to model a variable focal length",
+           "type": "object",
+           "properties": {
+              "key": {
```
jlaura: It is not clear what the available keys are for this.

schemas/isd.json
jlaura: :+1: I just implemented to package one of these and passing a scalar for say CTX would be ideal.

schemas/isd.json
```diff
@@ -0,0 +1,236 @@
...
+      },
+      "detector_line_summing": {
+         "description": "The summing in the line-direction",
+         "type": "number"
+      },
+      "apply_light_time_correction": {
+         "description": "A flag to indcate whether the light-time correction should be applied",
+         "type": "bool"
+      },
+      "number_of_ephem": {
```
jlaura: The spec is mixing 'quat' and 'quaternion' and 'ephem' and 'ephemeris'.  I would rather just use the full word to be very explicit, but consistency is more important than preference.

schemas/isd.json
```diff
@@ -0,0 +1,236 @@
...
+      "reference_height": {
+         "description": "Used for determining the intersection of the look vector with the ellipsoid",
+         "type": "array",
+         "items":  {
+            "type": "number",
+            "minItems": 2,
+            "maxItems": 2
+         }
+      },
+      "interpolation_method": {
```
jlaura: This is another instance where having an enum of available types would be a nice to have.

schemas/isd.json
kberryUSGS: :+1:

schemas/isd.json
kberryUSGS: :+1: Agreed. I'm unsure what the available options are. Do you know what the list of things to include here would be?

jessemapel: @twilson271828 Would be the person to go to on this.  I think the only method that is supported right now is lagrange

kberryUSGS: Is it useful to switch to using an enumeration for this that currently only has one valid value?

jessemapel: That sounds like a good idea to me

schemas/isd.json
kberryUSGS: I updated this to use the full words ephemeris (or ephemerides) and quaternion.

schemas/isd.json
kberryUSGS: I did go ahead and link to that NAIF whitepaper. Let me know if you ended up finding anything different / better to describe quaternion formats?

schemas/isd.json
kberryUSGS: :+1:

kberryUSGS:
Does anyone know why the AppVeyor build is failing with: `The batch file cannot be found.
122Command exited with code 1`

kberryUSGS:
This is updated on my fork but not yet pulled into Tyler's

kberryUSGS:
@jlaura @jessemapel @thareUSGS Travis is passing now. Would someone be willing to merge this?

PR #32, opened 2018-03-08
Changed ground to image iteration from Newton's method to bisection

### LINKED ISSUE #30: Generic line scan camera's ground to image iteration is very inefficient
The generic line scan camera uses Newton's method to calculate sample and line. Looking at some of the literature on this problem, that looks like a possibly poor method.

See:
https://www.asprs.org/wp-content/uploads/pers/1993journal/may/1993_may_655-661.pdf
https://www.asprs.org/wp-content/uploads/pers/2001journal/april/2001_apr_449-452.pdf
https://www.asprs.org/wp-content/uploads/pers/2009journal/september/2009_sep_1059-1067.pdf

We should at the very least switch to solving for line (down-track) and then sample (across-track). ISIS uses a bisecting window search method, that we should consider implementing.

jessemapel:
Resolves #30 
I also changed the line scan camera tests to use the CSM intersection values.  I left the ISIS3 values commented out above them for future reference.


jessemapel:
The new ground to image routine is based on pg42 here https://earth.esa.int/documents/700255/1929094/N77D7-PV02-US-5-IQC-GC-ATBD-ACS-0102-2.1/582f28ab-707c-4640-997c-8073d16d224f

jessemapel:
This also brings us in line with how ISIS calculates ground to image for line scan cameras

jessemapel:
The new ground to image method can take more iterations than the old Newton's method approach, but each iteration takes about 1/3 to 1/4 of the time.

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -200,156 +200,120 @@ csm::ImageCoord UsgsAstroLsSensorModel::groundToImage(
...
+                                           _data.m_IntTimeStartTimes.end(),
+                                           computedTime);
+   if (referenceTimeIt != _data.m_IntTimeStartTimes.begin()) {
+      --referenceTimeIt;
+   }
+   size_t referenceIndex = std::distance(_data.m_IntTimeStartTimes.begin(), referenceTimeIt);
+   calculatedPixel.line += _data.m_IntTimeLines[referenceIndex] - 1
+                         + (computedTime - _data.m_IntTimeStartTimes[referenceIndex])
+                         / _data.m_IntTimes[referenceIndex];
+   csm::EcefCoord calculatedPoint = imageToGround(calculatedPixel, _data.m_RefElevation);
```
jessemapel: Until #29 is merged in, this part may cause problems when the actual pixel location is close to 0 or 1 lines, because the imagetoground call here won't properly compute the ground point.  So, for HRSC the new ground point could be off by +-20 m. This will cause an error in the next section.

PR #54, opened 2018-08-01
Adds appveyor support

### LINKED ISSUE #52: Appveyor
Get the windows build working on appveyor.

jlaura:
Closes #52


jlaura:
@ihumphrey-usgs Let's keep this open until the appveyor CI passes. Then we can squash and merge (and get the setup ported over to the SensorUtils repo)

jlaura:
So that was epic. I would suggest that this is good to merge as the issue is calling for windows builds to be enabled. They are not passing...but the issue is not longer appveyor. I **think** that we need to include some `__declspec(dllimport)` and `__declspec(dllexport)` flags inside of the `ifdefs` at the top of the classes.  I also think that the BAE examples that we have seen might have some of these.

jlaura:
Link to the build: https://ci.appveyor.com/project/jlaura/csm-cameramodel/build/0.1.74/job/2dvmgmj2d2b8b36c

If this does get merged, please `quash and merge` since we do not need all of the bad commit messages in there.

jlaura:
This should not get merged until the travis PR is merged.  I can then update this PR with the changes to CMakeLists.txt in that PR and we should be good to go (with the caveat that appveyor is good, but the build itself is not).

jlaura:
When you go to use this locally, it is necessary to pull the gtest submodule:

`git submodule update --init --recursive` 

otherwise the `gtest` directory will be empty and the cmake config will fail.

Issue #47, opened 2018-07-31
Set up ctest


jessemapel:
We've tried gtest in the past and it didn't play nice with cmake. ctest should work better with cmake.

jlaura:
From what I am seeing GTest is the framework to write C++ tests and CTest is the test collector and executor.  It looks like CMake 3.10+ should play nicely with GTest for writing the tests and CTest for executing them. It looks like [`gtest_discover_tests`](https://blog.kitware.com/dynamic-google-test-discovery-in-cmake-3-10/) is a way to have rich verbosity during test execution.

jessemapel:
Closed by #56

PR #56, opened 2018-08-01
Added basic ctest/gtest framework

jessemapel:
There's only 3 tests right now, two of which are non-sense, but we should be able to add tests from here.


jessemapel:
See #47

CMakeLists.txt
```diff
@@ -14,19 +14,32 @@ add_library(usgscsm SHARED
...
+    UsgsAstroLsStateData.h
 )
 
 find_library(CSM_LIBRARY csmapi REQUIRED)
 find_path(CSM_INCLUDE_DIR NAMES csm)
 
-target_include_directories(usgscsm PRIVATE
-                           include/
+target_include_directories(usgscsm
+                           PUBLIC
```
jlaura: I think theses can stay PRIVATE?

CMakeLists.txt
```diff
@@ -14,19 +14,32 @@ add_library(usgscsm SHARED
...
     LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
-    PUBLIC_HEADER DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
\ No newline at end of file
+    PUBLIC_HEADER DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
+
+include(CTest)
+
+enable_testing()
+
+add_subdirectory(tests)
```
jlaura: 👍

PR #62, opened 2018-08-02
Add simple framer ISD for tests

### LINKED ISSUE #63: Create a simple fabricated “Pinhole” Framer ISD for test purposes
This includes simple rotation angles, location, etc

ihumphrey-usgs:
fixes #63 

Note that this is using the keywords from the Framer model we have, not keywords from the JSON schema (yet).

### DIFF
--- a/tests/simpleFramerISD.json
+++ b/tests/simpleFramerISD.json
@@ -20,7 +20,7 @@
     "nsamples": 5,
     "omega": 0,
     "phi": 0,
-    "semi_major_axis": 100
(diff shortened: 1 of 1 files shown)


tests/simpleFramerISD.json
```diff
@@ -0,0 +1,37 @@
+{
+ 
+    "ephemeris_time": 418855170.49264956,
```
jessemapel: Can we make this like 100?

PR #69, opened 2018-08-03
Adds CTest/GTest/Coveralls

jlaura:


.travis.yml
```diff
@@ -30,9 +30,14 @@ install:
   - bash miniconda.sh -b -p $HOME/miniconda
   - export PATH="$HOME/miniconda/bin:$PATH"
   - conda install -y -c conda-forge cmake
+  - pip install cpp-coveralls
```
jessemapel: Using conda and pip to manage packages seems odd. Can we just use one or the other?

tests/CMakeLists.txt
```diff
@@ -1,7 +1,15 @@
...
+target_link_libraries(runTests PUBLIC usgscsm ${GTEST_LIBRARIES} ${GTEST_MAIN_LIBRARIES} pthread)
+
+if (COVERAGE)
+  target_compile_options(runTests PRIVATE --coverage)
+  target_link_libraries(runTests PRIVATE --coverage)
+endif()
 
-gtest_discover_tests(runTests WORKING_DIRECTORY tests)
+gtest_discover_tests(runTests tests)
+message("Working dir: ${WORKING_DIRECTORY}")
```
jessemapel: Why not set the working directory? How do we ensure data file paths for testing specifying working directory?

.travis.yml
jlaura: Yes and no. I looked at this.  cpp-coveralls is available via conda, but not from a major provider. `pip` seemed to be a safer path because we know that pypi will have the dep.

tests/CMakeLists.txt
jlaura: I need to pull the extraneous message

PR #75, opened 2018-08-09
Adds conda deploy via travis

jlaura:
To discuss: with how cmake is building our shared library, the conda build also includes the libcsmapi.so is that what we want?


jessemapel:
I think it's not the end of the world as long as it plays okay with other installs of the csmapi by itself

jlaura:
@jessemapel Refactored this out to now optionally build the CSM. This way the libcsm is clean. Assuming that the CI passes - this is clean to complete in terms of feature additions/ changes. Tagging @makaylas to review.

PR #78, opened 2018-08-14
Improved frame plugin testing

jessemapel:
This is not exhaustive but it's a start.

### DIFF
--- a/tests/TestyMcTestFace.cpp
+++ b/tests/TestyMcTestFace.cpp
@@ -118,9 +118,6 @@ TEST_F(FrameIsdTest, ConstructInValidCamera) {
    UsgsAstroFramePlugin testPlugin;
    // Remove the model_name keyword from the ISD to make it invalid
    isd.clearParams("model_name");
-   std::string badState = "{"
-         "\"model_name\":\"USGS_ASTRO_FRAME_SENSOR_MODEL\","
-         "\"bad_param\":\"bad_value\"}";
    csm::Model *cameraModel;
    EXPECT_THROW(
          testPlugin.constructModelFromISD(
--- a/tests/TestyMcTestFace.cpp
+++ b/tests/TestyMcTestFace.cpp
@@ -18,18 +18,15 @@ class FrameIsdTest : public ::testing::Test {
    virtual void SetUp() {
       std::ifstream isdFile("data/simpleFramerISD.json");
       json jsonIsd = json::parse(isdFile);
-      isd.clearAllParams();
       for (json::iterator it = jsonIsd.begin(); it != jsonIsd.end(); ++it) {
-         if (it.value().size() >1 ) {
-            std::vector<double> v = it.value();
-            for (int j=0;j < v.size(); j++) {
-               std::ostringstream val;
-               val << std::setprecision(15) << v[j];
-               isd.addParam(it.key(),val.str());
+         json jsonValue = it.value();
+         if (jsonValue.size() > 1) {
+            for (int i = 0; i < jsonValue.size(); i++) {
+               isd.addParam(it.key(), jsonValue[i].dump());
             }
          }
          else {
-            isd.addParam(it.key(), it.value().dump());
+            isd.addParam(it.key(), jsonValue.dump());
          }
       }
       isdFile.close();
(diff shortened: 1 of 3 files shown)


tests/TestyMcTestFace.cpp
```diff
@@ -78,12 +91,48 @@ TEST(FramePluginTests, MissingStateValue) {
...
+   UsgsAstroFramePlugin testPlugin;
+   csm::Model *cameraModel;
+   EXPECT_NO_THROW(
+         cameraModel = testPlugin.constructModelFromISD(
+               isd,
+               "USGS_ASTRO_FRAME_SENSOR_MODEL",
+               NULL)
+   );
+   UsgsAstroFrameSensorModel *frameModel = dynamic_cast<UsgsAstroFrameSensorModel *>(cameraModel);
+   EXPECT_FALSE(frameModel == NULL);
```
jessemapel: There's gotta be a better way to check if a pointer is NULL in gtest, but this is the best I could find.

tests/TestyMcTestFace.cpp
```diff
@@ -78,12 +91,45 @@ TEST(FramePluginTests, MissingStateValue) {
...
+   UsgsAstroFramePlugin testPlugin;
+   // Remove the model_name keyword from the ISD to make it invalid
+   isd.clearParams("model_name");
+   csm::Model *cameraModel;
+   EXPECT_THROW(
+         testPlugin.constructModelFromISD(
+               isd,
+               "USGS_ASTRO_FRAME_SENSOR_MODEL",
+               NULL),
+         csm::Error
```
jessemapel: This is currently just checking the we caught a generic CSM error. I'd also like to check that it's the correct error type and the message is correct but I don't know how to do that in gtest

tests/TestyMcTestFace.cpp
jlaura: What about `IsNull`? https://github.com/google/googletest/blob/master/googlemock/docs/CheatSheet.md#generic-comparison

tests/TestyMcTestFace.cpp
```diff
@@ -18,8 +20,19 @@ class FrameIsdTest : public ::testing::Test {
       json jsonIsd = json::parse(isdFile);
       isd.clearAllParams();
```
jlaura: Is the setUp rerun each time and it is necessary to clear the ISD before a second run?

tests/TestyMcTestFace.cpp
```diff
@@ -18,8 +20,19 @@ class FrameIsdTest : public ::testing::Test {
...
       for (json::iterator it = jsonIsd.begin(); it != jsonIsd.end(); ++it) {
-         isd.addParam(it.key(), it.value().dump());
+         if (it.value().size() >1 ) {
+            std::vector<double> v = it.value();
+            for (int j=0;j < v.size(); j++) {
```
jlaura: I thought in the JSON library it is not necessary to cast this into a standard vector? https://github.com/nlohmann/json#stl-like-access

tests/TestyMcTestFace.cpp
```diff
@@ -18,8 +20,19 @@ class FrameIsdTest : public ::testing::Test {
...
+               std::ostringstream val;
+               val << std::setprecision(15) << v[j];
+               isd.addParam(it.key(),val.str());
```
jlaura: I would rather see precision being set in a single place - inside of the ISD.

tests/TestyMcTestFace.cpp
jlaura: https://stackoverflow.com/questions/23270078/test-a-specific-exception-type-is-thrown-and-the-exception-has-the-right-propert

tests/TestyMcTestFace.cpp
jessemapel: :+1:

tests/TestyMcTestFace.cpp
jessemapel: I was hoping for a better way to do that. I feel like there should be an approach that doesn't use the fail assertion.

jlaura:
I only caught the most recent change and approved - other comments are still inline.

tests/TestyMcTestFace.cpp
jessemapel: setUp is called before every test that uses the framework. We actually don't need to clear the ISD

tests/TestyMcTestFace.cpp
jessemapel: I think the precision can just be removed

jlaura:
@jessemapel This looks good to me. Are you happy with a merge?

PR #84, opened 2018-08-17
[DO NOT MERGE] Added ctx test and split up tests

jessemapel:
I'm putting this up mainly to get feedback on how we should split up tests across multiple files. I've set this up with one file that is just the gtest main and then multiple source files with tests for different things. TestyMcTestFace, long may he reign, should probably get renamed to FramePluginTests or something.


jessemapel:
Now that Kaj's test are in I'm going to close this without merging

PR #87, opened 2018-08-21
Updated ISD and tests 

Kelvinrr:
Camera models from the ISD are being created, but the camera model is still failing, I imagine because some functions still need to be changed to accommodate the new changes (quats + maybe unit conversion issues).  Probably have to wait for Tuesday, enough here to review by tomorrow morning. 

Chunks left commented out are parts that I am unsure about whether or not a ISD element belongs there. If no one tells me otherwise, I'll just delete them before the next changes.


jlaura:
:+1: I am going to pull this locally to see if I can make the changes necessary.  I think I know where the issue is.

tests/data/simpleFramerISD.json
```diff
@@ -1,96 +1,53 @@
 {
```
jlaura: I had not noticed, but this is one cause of the failing tests.  We need to maintain the simple ISD (that was removed). The more complex one is 'too' complex for the 'simle' testing being done.

tests/data/simpleFramerISD.json
```diff
@@ -1,96 +1,53 @@
...
+  "image_samples": 1024,
+  "interpolation_method": "lagrange",
+  "number_of_ephemerides": 1,
+  "optical_distortion": {
+    "x": [0.0, 1.001854269623802, 0.0, 0.0, -0.0005094440474941111, 0.0, 1.004010471468856e-05, 0.0, 1.004010471468856e-05, 0.0],
+    "y": [0.0, 0.0, 1.0, 0.0009060010594996751, 0.0, 0.0003574842626620758, 0.0, 1.004010471468856e-05, 0.0, 1.004010471468856e-05]
+  },
+  "radii": {
+    "semimajor": 2439400.0,
+    "semiminor": 2439400.0
```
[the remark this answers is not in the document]
Kelvinrr: probably should, swagger spec doesn't have the units for Radii, will add in.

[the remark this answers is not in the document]
jlaura: :+1: I am working on the opk<->quaternion stuff now. I updated the rest of the spec with the simple values.

tests/Fixtures.h
```diff
@@ -0,0 +1,76 @@
...
+#include <json.hpp>
+
+#include <sstream>
+#include <fstream>
+
+#include <gtest/gtest.h>
+
+using json = nlohmann::json;
+
+// Should this be positioned somehwere in the public API?
```
jessemapel: The public API is the CSM API so we can't put utility functions like this out in this library. This will have to go into a utility library like the SensorUtils library.

src/UsgsAstroFramePlugin.cpp
```diff
@@ -277,207 +273,199 @@ csm::Model *UsgsAstroFramePlugin::constructModelFromISD(const csm::Isd &imageSup
...
-  if (imageSupportData.param("transx", 0) == "") {
-    missingKeywords.push_back("transx 0");
-  }
-  else if (imageSupportData.param("transx", 1) == "") {
-    missingKeywords.push_back("transx 1");
-  }
-  else if (imageSupportData.param("transx", 2) == "") {
-    missingKeywords.push_back("transx");
+  if (imageSupportData.param("sensor_orientation", 3) == "") {
+    missingKeywords.push_back("kappa");
```
jessemapel: The missing keywords for Omega, Phi, Kappa should be changed to just check if sensor orientation has a quaternion, 4 elements. If it doesn't, it should append "sensor_orientation" to the missingKeywords.

tests/Fixtures.h
```diff
@@ -0,0 +1,76 @@
...
+inline void jsonToIsd(json &object, csm::Isd &isd, std::string prefix="") {
+  for (json::iterator it = object.begin(); it != object.end(); ++it) {
+     json jsonValue = it.value();
+
+     // Is this type of type sensative de-serialization not available
+     // out of the box?
+     if (jsonValue.is_object()) {
+       // Is flattening the best approach? not sure...
+       jsonToIsd(jsonValue, isd, prefix+it.key()+'_');
```
jessemapel: I don't think we want to flatten the JSON. What we can do is convert the object into a JSON encoded string and then just pass that as the ISD parameter value. The plugin can then parse out the object.

tests/Fixtures.h
```diff
@@ -0,0 +1,76 @@
...
+
+using json = nlohmann::json;
+
+// Should this be positioned somehwere in the public API?
+inline void jsonToIsd(json &object, csm::Isd &isd, std::string prefix="") {
+  for (json::iterator it = object.begin(); it != object.end(); ++it) {
+     json jsonValue = it.value();
+
+     // Is this type of type sensative de-serialization not available
+     // out of the box?
```
jessemapel: I think we can define some stuff to make this work more elegantly, the JSON library is rather complex because of how much it does. I've tried this before and it was a huge pain debugging it.

tests/Fixtures.h
```diff
@@ -0,0 +1,76 @@
+#pragma once
```
jessemapel: We generally use define include guards instead of `#pragma once`

jessemapel: I'm open to using `#pragma once` but we should be consistent through out the library, all of the other headers define `FILENAME_H`

tests/data/simpleFramerISD.json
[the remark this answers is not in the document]
jessemapel: The plugin assumes radii unites are km. I can't remember if the ISD spec says meters or km here. If it's meters, then we need to change that behavior in the plugin.

src/UsgsAstroFramePlugin.cpp
```diff
@@ -277,207 +273,199 @@ csm::Model *UsgsAstroFramePlugin::constructModelFromISD(const csm::Isd &imageSup
...
+    missingKeywords.push_back("focal2pixel_samples 0");
+  }
+  else if (imageSupportData.param("focal2pixel_samples", 1) == "") {
+    missingKeywords.push_back("focal2pixel_samples 1");
+  }
+  else if (imageSupportData.param("focal2pixel_samples", 2) == "") {
+    missingKeywords.push_back("focal2pixel_samples 2");
+  }
+
+  sensorModel->m_majorAxis = 1000 * atof(imageSupportData.param("radii_semimajor").c_str());
```
jessemapel: Implicit conversion of km radii to meters here.

src/UsgsAstroFramePlugin.cpp
```diff
@@ -277,207 +273,199 @@ csm::Model *UsgsAstroFramePlugin::constructModelFromISD(const csm::Isd &imageSup
...
+    missingKeywords.push_back("radii_semimajor");
   }
   // Do we assume that if we do not have a semi-minor axis, then the body is a sphere?
-  if (imageSupportData.param("semi_minor_axis") == "") {
+  if (imageSupportData.param("radii_semiminor") == "") {
     sensorModel->m_minorAxis = sensorModel->m_majorAxis;
   }
   else {
-    sensorModel->m_minorAxis = 1000 * atof(imageSupportData.param("semi_minor_axis").c_str());
+    sensorModel->m_minorAxis = 1000 * atof(imageSupportData.param("radii_semiminor").c_str());
```
jessemapel: Implicit conversion of km radii to meters here.

tests/data/simpleFramerISD.json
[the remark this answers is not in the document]
Kelvinrr: A lot of the implicit stuff is from the original plugin code, I could whip up a quick `toMeters` function that actually uses the units label and transforms everything to meters. Test ISD is already in new format, for the most part.

src/UsgsAstroFrameSensorModel.cpp
```diff
@@ -15,14 +15,15 @@ using namespace std;
...
   "Y Sensor Position (m)",  // 1
   "Z Sensor Position (m)",  // 2
-  "Omega (radians)",        // 3
-  "Phi (radians)",          // 4
-  "Kappa (radians)"         // 5
+  "w",        // 3
+  "x",        // 4
+  "y"         // 5
+  "z"         // 5
 };
```
jlaura: These should be v1, v2, v3 for x, y, z, respectively.

src/UsgsAstroFrameSensorModel.cpp
```diff
@@ -97,33 +98,33 @@ UsgsAstroFrameSensorModel::UsgsAstroFrameSensorModel() {
...
   m_iTransS[2] = 0.0;
 
   m_iTransL[0] = 0.0;
   m_iTransL[1] = 0.0;
-  m_iTransL[2] = 0.0;
+  m_iTransL[2] = 10.0;
 
   m_ephemerisTime = 0.0;
-  m_originalHalfSamples = 0.0;
+  m_originalHalfSamples = 8.0;
```
jlaura: should be 0.0

src/UsgsAstroFrameSensorModel.cpp
```diff
@@ -97,33 +98,33 @@ UsgsAstroFrameSensorModel::UsgsAstroFrameSensorModel() {
...
+  m_ccdCenter[1] = 7.5;
 
   m_line_pp = 0.0;
   m_sample_pp = 0.0;
 
   m_odtX.assign(10, 0.0);
   m_odtY.assign(10, 0.0);
 
-  m_originalHalfLines = 0.0;
+  m_originalHalfLines = 8.0;
```
jlaura: should be 0.0 - > for all of these changes, we want to keep the defaults as 0.0

Kelvinrr:
FYI I disabled appveyor in that it passes by default, from my understanding during a previous standup, it's still broken and needs to be fixed so I figure disabling it until then isn't a terrible idea. If anyone thinks otherwise I'll reverse it.

tests/data/simpleFramerISD.json
jlaura: This should be:

```json
{
	"center_ephemeris_time": 50,
	"dt_ephemeris": 100,
	"focal2pixel_lines": [0.0, 0.0, 10],
	"focal2pixel_samples": [0.0, 10, 0.0],
	"focal_length_model": {
		"focal_length": 50,
		"focal_length_epsilon": 1.0
	},
	"image_lines": 15,
	"image_samples": 15,
	"interpolation_method": "lagrange",
	"number_of_ephemerides": 1,
	"optical_distortion": {
		"x": [0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0],
		"y": [0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
	},
	"radii": {
		"semimajor": 0.01,
        "semiminor": 0.01,
        "unit":"m"
	},
	"reference_height": {
		"maxheight": 1,
		"minheight": -1,
		"unit": "m"
	},
	"sensor_location": {
		"unit": "m",
		"x": 1000,
		"y": 0,
		"z": 0
	},
	"sensor_orientation": [0.0, -0.707106781186547, 0.0, 0.707106781186547],
	"sensor_velocity": {
		"unit": "m",
		"x": 0,
		"y": 0,
		"z": 0
	},
	"starting_detector_line": 0,
	"starting_detector_sample": 0,
	"starting_ephemeris_time": 0,
	"sun_position": {
		"x": 100,
		"y": 100,
		"z": 0
	},
	"sun_velocity": {
		"x": 0,
		"y": 0,
		"z": 0
	}
}

I need to checkout the plugin re: a comment from @jessemaple but this ISD should be good.
```

src/UsgsAstroFrameSensorModel.cpp
jlaura: This also means that the pulling of these in the calculation of the rotation matrices should probably use the proper variable names. Instead of the lazy x,y,z

jlaura:
re: appveyor :+1:

src/UsgsAstroFrameSensorModel.cpp
Kelvinrr: These changes were only because image2ground relies on params not in the ISD anymore. I imagine the thing to do is reset them and let the tests fail and in a separate PR, someone else can make those changes.

src/UsgsAstroFrameSensorModel.cpp
```diff
@@ -97,33 +98,33 @@ UsgsAstroFrameSensorModel::UsgsAstroFrameSensorModel() {
...
   m_spacecraftName = "";
-  m_pixelPitch = 0.0;
+  m_pixelPitch = 0.1;
```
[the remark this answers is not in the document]
Kelvinrr: see response put onto Jay's comment.

Kelvinrr:
I believe I addressed all the comments. Because the defaults are all 0, the tests are going to fail. There are parameters that `imageToGround` relies on that are no longer in the ISD. I think the best approach is to merge with the failed tests and someone else in a different PR can change some of the sensor model functions to match to the new ISD and can update tests appropriately.

tests/Fixtures.h
Kelvinrr: I'm more biased to pragma once, but I think there enough changes here for now.

src/UsgsAstroFramePlugin.cpp
```diff
@@ -293,191 +289,139 @@ csm::Model *UsgsAstroFramePlugin::constructModelFromISD(const csm::Isd &imageSup
...
+  sensorModel->m_odtY[0] = atof(imageSupportData.param("optical_distortion_y", 0).c_str());
+  sensorModel->m_odtY[1] = atof(imageSupportData.param("optical_distortion_y", 1).c_str());
+  sensorModel->m_odtY[2] = atof(imageSupportData.param("optical_distortion_y", 2).c_str());
+  sensorModel->m_odtY[3] = atof(imageSupportData.param("optical_distortion_y", 3).c_str());
+  sensorModel->m_odtY[4] = atof(imageSupportData.param("optical_distortion_y", 4).c_str());
+  sensorModel->m_odtY[5] = atof(imageSupportData.param("optical_distortion_y", 5).c_str());
+  sensorModel->m_odtY[6] = atof(imageSupportData.param("optical_distortion_y", 6).c_str());
+  sensorModel->m_odtY[7] = atof(imageSupportData.param("optical_distortion_y", 7).c_str());
+  sensorModel->m_odtY[8] = atof(imageSupportData.param("optical_distortion_y", 8).c_str());
+  sensorModel->m_odtY[9] = atof(imageSupportData.param("optical_distortion_y", 9).c_str());
```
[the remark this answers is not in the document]
Kelvinrr: not entirely sure what you are asking. I don't know very much about the distortion models, so you need to be more specific. like what should be the expectation for arrays whose length  is < 10. pad with zeroes? because that's the current behavior. If not, what else?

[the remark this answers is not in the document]
jlaura: Disclaimer: This could be the wrong way to do this.
I coded the state stuff to utilize vectors. I wonder if the distortion could do that? If we assume all of the coefficients are floats, I think we can just use the standard type? This also plays nice with the JSON lib.

https://github.com/USGS-Astrogeology/CSM-CameraModel/blob/master/src/UsgsAstroFramePlugin.cpp#L266

Kelvinrr:
@jlaura I would vote towards keeping this quarantined until it's ready. A lot of things need to be changed all at once and I don't think one monster PR is the way to go. I doubt some of the changes in this PR are final, and there is still more to go.

Kelvinrr:
@jlaura someone with Admin privileges needs to make the remote branch before I can change into it. maybe just a dev branch?

jlaura:
@Kelvinrr dev branch created.

src/UsgsAstroFramePlugin.cpp
```diff
@@ -277,207 +259,280 @@ csm::Model *UsgsAstroFramePlugin::constructModelFromISD(const csm::Isd &imageSup
...
-  if (imageSupportData.param("ephemeris_time") == "") {
-    missingKeywords.push_back("ephemeris_time");
+  sensorModel->m_nLines = atoi(imageSupportData.param("image_lines").c_str());
+  sensorModel->m_nSamples = atoi(imageSupportData.param("image_samples").c_str());
+  if (imageSupportData.param("image_lines") == "") {
+    missingKeywords.push_back("image_lines");
+  }
+  if (imageSupportData.param("image_samples") == "") {
+    missingKeywords.push_back("image_samples");
   }
```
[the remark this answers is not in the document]
Kelvinrr: If still in the scope of this PR, just tell me where they need to go and how to acquire the principle point and I can slap that in real quick.

jlaura:
@Kelvinrr I think fine to omit for now, we have a number of these 'derived' members that should probably all get lumped into a different PR.

src/UsgsAstroFramePlugin.cpp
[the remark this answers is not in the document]
jessemapel: The `detector_center` parameter determines the center of the ccd. We currently don't have a way to pass a principal point offset. That could be included in the distortion specification.

PR #88, opened 2018-08-21
Updates #87

jlaura:
I believe that the computation of the rotation matrix from the quaternion is correct. I used the homogeneous expression from [here](https://en.wikipedia.org/wiki/Conversion_between_quaternions_and_Euler_angles#Rotation_matrices).

When converting from the euler angles, I think that my quaternion ordering (in the simpleFramerISD.json) is incorrect. Therefore the tests are still failing.  I need to look at fixing the quaternion ordering. Another set of eyes appreciated.

Great work from @Kelvinrr getting all of the ISD keywords swapped in.

### DIFF
--- a/src/UsgsAstroFramePlugin.cpp
+++ b/src/UsgsAstroFramePlugin.cpp
@@ -10,7 +10,7 @@
 #include <Warning.h>
 #include <Version.h>
 
-#include <json/json.hpp>
+#include <json.hpp>
 
 
 using json = nlohmann::json;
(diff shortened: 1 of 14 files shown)


jlaura:
@Kelvinrr if I can get the tests passing in here, we can get this merged back into your open PR and then I will close this.

src/UsgsAstroFramePlugin.cpp
```diff
@@ -277,207 +273,199 @@ csm::Model *UsgsAstroFramePlugin::constructModelFromISD(const csm::Isd &imageSup
...
+    missingKeywords.push_back("w");
+  }
+  if (imageSupportData.param("sensor_orientation", 1)== "") {
+    missingKeywords.push_back("v1");
+  }
+  if (imageSupportData.param("sensor_orientation", 2) == "") {
+    missingKeywords.push_back("v2");
+  }
+  if (imageSupportData.param("sensor_orientation", 3) == "") {
+    missingKeywords.push_back("v3");
```
jessemapel: Our warnings about what failed in ISD construction should use the keywords from the ISD spec. We should either report an error with "sensor_orientation" or make the quaternions an object so that "w", "v1", "v2", and "v3" are meaningful to users.

src/UsgsAstroFrameSensorModel.cpp
```diff
@@ -15,14 +15,15 @@ using namespace std;
...
   "X Sensor Position (m)",  // 0
   "Y Sensor Position (m)",  // 1
   "Z Sensor Position (m)",  // 2
-  "Omega (radians)",        // 3
-  "Phi (radians)",          // 4
-  "Kappa (radians)"         // 5
+  "w",                      // 3
+  "x",                     // 4
+  "y",                     // 5
+  "z"                      // 6
```
jessemapel: Should x, y, z be v1, v2, v3?

tests/data/simpleFramerISD.json
```diff
@@ -1,96 +1,54 @@
...
+		"x": 1000,
+		"y": 0,
+		"z": 0
+	},
+	"sensor_orientation": [0.0, -0.707106781186547, 0.0, 0.707106781186547],
+	"sensor_velocity": {
+		"unit": "m",
+		"x": 1,
+		"y": 1,
+		"z": 1
```
jessemapel: We should use 0 velocity in this basic test.

jessemapel: Testing that velocity is properly accounted for is a separate test.

jessemapel:
quaternions and rotation matrix look good to me

tests/data/simpleFramerISD.json
```diff
@@ -1,96 +1,54 @@
...
+	"interpolation_method": "lagrange",
+	"number_of_ephemerides": 1,
+	"optical_distortion": {
+		"x": [0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0],
+		"y": [0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
+	},
+	"radii": {
+		"semimajor": 0.01,
+        "semiminor": 0.01,
+        "unit":"KM"
```
thareUSGS: switch to meters

jessemapel:
Guess at why tests are failing, the new ISD doesn't specify where the origin of the camera reference frame is. This can be baked into the focal2pixels parameters

tests/data/simpleFramerISD.json
```diff
@@ -1,96 +1,54 @@
...
-        0.0,
-        0.0
-    ],
-    "starting_detector_line": 0.0,
-    "starting_detector_sample": 0.0
-}
+	"center_ephemeris_time": 50,
+	"dt_ephemeris": 100,
+	"focal2pixel_lines": [0.0, 0.0, 0.1],
+	"focal2pixel_samples": [0.0, 0.1, 0.0],
```
jessemapel: These parameters should be focal plane x, y to pixel conversions. These values are pixels to focal plane x, y

jessemapel: In the plugin these values are being put in the pixel to focal plane coefficient vectors, so it's a double negative situation.

src/UsgsAstroFramePlugin.cpp
```diff
@@ -277,207 +273,199 @@ csm::Model *UsgsAstroFramePlugin::constructModelFromISD(const csm::Isd &imageSup
...
+  if (imageSupportData.param("image_lines") == "") {
+    missingKeywords.push_back("image_lines");
+  }
+  if (imageSupportData.param("image_samples") == "") {
+    missingKeywords.push_back("image_samples");
+  }
+
+  sensorModel->m_transY[0] = atof(imageSupportData.param("focal2pixel_lines", 0).c_str());
+  sensorModel->m_transY[1] = atof(imageSupportData.param("focal2pixel_lines", 1).c_str());
+  sensorModel->m_transY[2] = atof(imageSupportData.param("focal2pixel_lines", 2).c_str());
```
jessemapel: m_transY is the coefficients for the affine from pixels to focal plane y. The focal2pixel parameters should be the inverse of that. We'll have to modify the camera model to make this work correctly. Right now, we have the wrong parameters in `focal2pixel_*` so it's not causing an error, but both should be fixed.

jlaura:
@jessemapel comments moved into #87 where @Kelvinrr has made a bunch of these changes. I am checking out the plugin for conversions (trans, itrans)

PR #89, opened 2018-08-22
Added framing test for setFocalPlane.  

twilson271828:
I also added the ability to test protected/private member functions in classes using the FRIEND_TEST macro in gtest (although I am not happy with this particular implementation).

Additionally in the test fixture, I load the ISD and replace the odt_x,odt_y vectors with those defined within
the ik kernel for the MDIS NAC.  I did this through the ISD.  There might be a better way to do it
through the nlohmann::json api though.

### DIFF
--- a/tests/FrameCameraTests.cpp
+++ b/tests/FrameCameraTests.cpp
@@ -266,7 +266,7 @@ TEST_F(FrameIsdTest, distortMe_AllCoefficientsOne) {
    UsgsAstroFrameSensorModel* sensorModel = dynamic_cast<UsgsAstroFrameSensorModel *>(model);
 
 
-
+   printIsd(isd);
    double dx,dy;
    sensorModel->distortionFunction(imagePt.samp, imagePt.line,dx,dy );
 
@@ -274,7 +274,7 @@ TEST_F(FrameIsdTest, distortMe_AllCoefficientsOne) {
    cout << "dy:  " << dy << endl;
 
    EXPECT_NEAR(dx,1872.25,1e-8 );
-   EXPECT_NEAR(dy,843.75,1e-8);
+   EXPECT_NEAR(dy,1872.25,1e-8);
 
 }
 
@@ -288,7 +288,7 @@ TEST_F(FrameIsdTest, setFocalPlane_AllCoefficientsOne) {
   isd.clearParams(odty_key);
   isd.clearParams(odtx_key);
 
-   csm::ImageCoord imagePt(1872.25, 843.75);
+   csm::ImageCoord imagePt(1872.25, 1872.25);
 
    vector<double> odtx{1.0,1.0,1.0,1.0,1.0,1.0,1.0,1.0,1.0,1.0};
    vector<double> odty{1.0,1.0,1.0,1.0,1.0,1.0,1.0,1.0,1.0,1.0};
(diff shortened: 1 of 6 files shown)


jlaura:
Second to not using real values. If we end up with a corner case that requires real (or pseudo-real) values, we should take a long look at why this is the case. At this point, we need to be doing tests on paper using simple values to ensure that we are 100% correct. The problem with real values is that we are (hopefully) going to start to build on top of these methods and we need them to be bomber.

twilson271828:
I'll try that (modifying the ISD fixture).  The only way to test actual distortion in this case would be to use an actual MDIS NAC image.

jessemapel:
@twilson271828 We can use the test "image" and introduce artificial distortion. That will test without a real image

twilson271828:
So any comments about the way I modified the testing to test protected/private class members?

jessemapel:
@twilson271828 The private member testing looks good. The gtest docs recommend not using the private testing tools and refactor, but because of the CSM API we can't. So, I think this is the best way to test those types of functions.

tests/FrameCameraTests.cpp
```diff
@@ -101,3 +129,32 @@ TEST_F(FrameSensorModel, OffBody4) {
...
+      strval << setprecision(precision) << val;
+      localIsd.addParam("odt_x", strval.str());
+   }
+   for (auto & val: odty){
+      ostringstream strval;
+      strval << setprecision(precision) << val;
+      localIsd.addParam("odt_y", strval.str());
+   }
+
+   sensorModel->setFocalPlane(imagePt.samp, imagePt.line, ux, uy);
```
jessemapel: The sensor model is not updated from your local ISD.  You need to modify the ISD and then generate a camera model from it. Look at how @AstroKEW78 wrote the focal length test https://github.com/USGS-Astrogeology/CSM-CameraModel/blob/master/tests/FrameCameraTests.cpp#L127

tests/FrameCameraTests.cpp
```diff
@@ -124,20 +147,51 @@ TEST_F(FrameSensorModel, OffBody4) {
...
+   EXPECT_NEAR(ux,7.0,1e-8 );
+   EXPECT_NEAR(uy,7.0,1e-8);
```
jessemapel: Shouldn't these by 14,63 ?

tests/FrameCameraTests.cpp
```diff
@@ -124,20 +147,51 @@ TEST_F(FrameSensorModel, OffBody4) {
...
 
+TEST_F(FrameSensorModel, setFocalPlane1) {
+  int precision  = 20;
+
+   csm::Isd localIsd;
+   loadIsd("data/simpleFramerISD.json",localIsd);
+   csm::ImageCoord imagePt(7.0, 7.0);
+   double ux,uy;
+   vector<double> odtx{0.0,1.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0};
+   vector<double> odty{0.0,1.0,1.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0};
```
jessemapel: All 0 vectors should be no distortion. This should just change one coefficient for x and one coefficient for y. Maybe just one coefficient at a time. This thing can be super parameterized.

tests/FrameCameraTests.cpp
```diff
@@ -124,20 +143,201 @@ TEST_F(FrameSensorModel, OffBody4) {
...
+   }
+
+   csm::Model *model = frameCameraPlugin.constructModelFromISD(
+         isd,
+         "USGS_ASTRO_FRAME_SENSOR_MODEL");
+
+   UsgsAstroFrameSensorModel* sensorModel = dynamic_cast<UsgsAstroFrameSensorModel *>(model);
+
+
+   printIsd(isd);
```
jessemapel: I can see this is helpful for debugging but I don't think we want to actually have this called as part of our test.

PR #91, opened 2018-08-22
removed 0.5 pixel offsets

AstroKEW78:


jessemapel:
@AstroKEW78 This change is going to break all the tests. You need to update the line, sample values in the tests to account for this.

PR #94, opened 2018-08-22
patched segfault error + added detector center as defined in json schema

Kelvinrr:
One last edit, noticed the last test failed due to a segfault, in hastily scanning the plugin member vars for things that should be 0 I accidentally passed 0.0 to std::vecotor::assign which should be a 10, essentially creating a 0 element vector.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ matrix:
         - MATRIX_EVAL="CXX=g++-6 && CC=gcc-6"
     - os: osx
       osx_image: xcode9.4
-      env: 
+      env:
         - MATRIX_EVAL="CXX=clang++ && CC=clang"
 
 before_install:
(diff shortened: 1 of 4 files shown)


jessemapel:
I'm not sure why this is failing CI. Can you add the -VV flag to the ctest call in the travis file?

jessemapel:
@Kelvinrr Nvm, just pull from master it has the -VV flag now.

jessemapel:
Actually, this isn't even expected to pass tests still so I think it's ready to merge?

Kelvinrr:
it's failing for the same reason I mentioned in the previous PR: there are some variables not defined in the ISD that imageToGround relies on and so they stay at 0, therefore the computed values are different.

jessemapel:
Okay, this is ready to go then.

PR #101, opened 2018-08-24
A few more focal plane tests

twilson271828:
I also added a createModel(*csm::Isd &isd) function which returns a pointer to a sensor model in the FrameIsdTest fixture, and then pulled out all of the copy-pasta in the tests Kaj and myself have been working on.

### DIFF
--- a/src/UsgsAstroFrameSensorModel.cpp
+++ b/src/UsgsAstroFrameSensorModel.cpp
@@ -206,8 +206,8 @@ csm::ImageCoord UsgsAstroFrameSensorModel::groundToImage(
 
   //Convert distorted mm into line/sample
   double sample, line;
-  sample = m_iTransS[0] + m_iTransS[1] * distortedx + m_iTransS[2] * distortedx + m_ccdCenter[0] - 0.5;
-  line =   m_iTransL[0] + m_iTransL[1] * distortedy + m_iTransL[2] * distortedy + m_ccdCenter[0] - 0.5;
+  sample = m_iTransS[0] + m_iTransS[1] * distortedx + m_iTransS[2] * distortedx + m_ccdCenter[0];
+  line =   m_iTransL[0] + m_iTransL[1] * distortedy + m_iTransL[2] * distortedy + m_ccdCenter[0];
 
   return csm::ImageCoord(line, sample);
 }
@@ -244,8 +244,8 @@ csm::EcefCoord UsgsAstroFrameSensorModel::imageToGround(const csm::ImageCoord &i
 
   //Convert from the pixel space into the metric space
   double optical_center_x, optical_center_y, x_camera, y_camera;
-  optical_center_x = m_ccdCenter[0] - 0.5;
-  optical_center_y = m_ccdCenter[1] - 0.5;
+  optical_center_x = m_ccdCenter[0];
+  optical_center_y = m_ccdCenter[1];
   y_camera = m_transY[0] + m_transY[1] * (lo - optical_center_y) + m_transY[2] * (lo - optical_center_y);
   x_camera = m_transX[0] + m_transX[1] * (so - optical_center_x) + m_transX[2] * (so - optical_center_x);
 
@@ -301,8 +301,8 @@ csm::EcefLocus UsgsAstroFrameSensorModel::imageToRemoteImagingLocus(const csm::I
                                                              csm::WarningList *warnings) const {
   // Find the line,sample on the focal plane (mm)
   // CSM center = 0.5, MDIS IK center = 1.0
-  double col = imagePt.samp - (m_ccdCenter[0] - 0.5);
-  double row = imagePt.line - (m_ccdCenter[1] - 0.5);
+  double col = imagePt.samp - (m_ccdCenter[0]);
+  double row = imagePt.line - (m_ccdCenter[1]);
   double focalPlaneX = m_transX[0] + m_transX[1] * col + m_transX[2] * col;
   double focalPlaneY = m_transY[0] + m_transY[1] * row + m_transY[2] * row;
 
--- a/src/UsgsAstroFrameSensorModel.cpp
+++ b/src/UsgsAstroFrameSensorModel.cpp
@@ -1043,10 +1043,9 @@ bool UsgsAstroFrameSensorModel::setFocalPlane(double dx,double dy,
                                        double &undistortedX,
                                        double &undistortedY ) const {
 
-
   // Solve the distortion equation using the Newton-Raphson method.
   // Set the error tolerance to about one millionth of a NAC pixel.
-  const double tol = 1.4E-5;
+  const double tol = 1.4E-7;
 
   // The maximum number of iterations of the Newton-Raphson method.
   const int maxTries = 60;
@@ -1066,7 +1065,8 @@ bool UsgsAstroFrameSensorModel::setFocalPlane(double dx,double dy,
 
   distortionFunction(x, y, fx, fy);
 
-  for (int count = 1; ((fabs(fx) + fabs(fy)) > tol) && (count < maxTries); count++) {
+
+  for (int count = 1; count < maxTries; count++) {
 
     this->distortionFunction(x, y, fx, fy);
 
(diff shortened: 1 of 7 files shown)


tests/FrameCameraTests.cpp
```diff
@@ -259,20 +264,13 @@ TEST_F(FrameIsdTest, distortMe_AllCoefficientsOne) {
...
-         isd,
-         "USGS_ASTRO_FRAME_SENSOR_MODEL");
-
-   UsgsAstroFrameSensorModel* sensorModel = dynamic_cast<UsgsAstroFrameSensorModel *>(model);
 
+   UsgsAstroFrameSensorModel* sensorModel = createModel(isd);
 
-   printIsd(isd);
    double dx,dy;
+   ASSERT_NE(sensorModel, nullptr);
```
jessemapel: also needs to clean up

tests/FrameCameraTests.cpp
```diff
@@ -304,14 +301,13 @@ TEST_F(FrameIsdTest, setFocalPlane_AllCoefficientsOne) {
...
-         isd,
-         "USGS_ASTRO_FRAME_SENSOR_MODEL");
 
-   UsgsAstroFrameSensorModel* sensorModel = dynamic_cast<UsgsAstroFrameSensorModel *>(model);
+
+   UsgsAstroFrameSensorModel* sensorModel = createModel(isd);
 
 
    double ux,uy;
+   ASSERT_NE(sensorModel, nullptr);
```
jessemapel: more sensor model clean up

tests/FrameCameraTests.cpp
```diff
@@ -323,20 +319,93 @@ TEST_F(FrameIsdTest, setFocalPlane_AllCoefficientsOne) {
...
+   }
+
+   UsgsAstroFrameSensorModel* sensorModel = createModel(isd);
+
+   double dx,dy;
+   ASSERT_NE(sensorModel, nullptr);
+   sensorModel->distortionFunction(imagePt.samp, imagePt.line,dx,dy );
+
+   EXPECT_NEAR(dx,908.5,1e-8 );
+   EXPECT_NEAR(dy,963.75,1e-8);
```
jessemapel: needs to be cleaned up too.

tests/FrameCameraTests.cpp
```diff
@@ -323,20 +319,93 @@ TEST_F(FrameIsdTest, setFocalPlane_AllCoefficientsOne) {
...
+
+   UsgsAstroFrameSensorModel* sensorModel = createModel(isd);
+
+   double ux,uy;
+   ASSERT_NE(sensorModel, nullptr);
+
+   sensorModel->setFocalPlane(imagePt.samp, imagePt.line,ux,uy );
+
+   EXPECT_NEAR(ux,7.5,1e-8 );
+   EXPECT_NEAR(uy,7.5,1e-8);
```
jessemapel: clean up the sensor model

src/UsgsAstroFrameSensorModel.cpp
```diff
@@ -1066,7 +1065,8 @@ bool UsgsAstroFrameSensorModel::setFocalPlane(double dx,double dy,
 
   distortionFunction(x, y, fx, fy);
 
-  for (int count = 1; ((fabs(fx) + fabs(fy)) > tol) && (count < maxTries); count++) {
+
+  for (int count = 1; count < maxTries; count++) {
```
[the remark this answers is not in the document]
twilson271828: Good point.  I don't know why I did that.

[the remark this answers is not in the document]
twilson271828: Fixed it.

tests/FrameCameraTests.cpp
```diff
@@ -24,9 +24,24 @@ class FrameIsdTest : public ::testing::Test {
...
+        UsgsAstroFramePlugin frameCameraPlugin;
+        csm::Model *model = frameCameraPlugin.constructModelFromISD(
+              modifiedIsd,"USGS_ASTRO_FRAME_SENSOR_MODEL");
+
+        UsgsAstroFrameSensorModel* sensorModel = dynamic_cast<UsgsAstroFrameSensorModel *>(model);
+
+        if (sensorModel)
+          return sensorModel;
+        else
+          return nullptr;
```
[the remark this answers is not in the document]
twilson271828: If you put an ASSERT_NE macro there, gcc throws  'error: void value not ignored as it ought to be.'

tests/FrameCameraTests.cpp
```diff
@@ -171,14 +185,8 @@ TEST_F(FrameIsdTest, setFocalPlane1) {
...
-
-
-
-   UsgsAstroFrameSensorModel* sensorModel = dynamic_cast<UsgsAstroFrameSensorModel *>(model);
-
+   UsgsAstroFrameSensorModel* sensorModel =createModel(isd);
+   ASSERT_NE(sensorModel, nullptr);
    sensorModel->setFocalPlane(imagePt.samp, imagePt.line, ux, uy);
    EXPECT_NEAR(imagePt.samp,7.5,1e-8 );
    EXPECT_NEAR(imagePt.line,7.5,1e-8);
```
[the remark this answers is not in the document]
twilson271828: Fixed it.

tests/FrameCameraTests.cpp
```diff
@@ -213,14 +220,12 @@ TEST_F(FrameIsdTest, Jacobian1) {
...
-         "USGS_ASTRO_FRAME_SENSOR_MODEL");
-
-   UsgsAstroFrameSensorModel* sensorModel = dynamic_cast<UsgsAstroFrameSensorModel *>(model);
+   UsgsAstroFrameSensorModel* sensorModel = createModel(isd);
 
 
    double Jxx,Jxy,Jyx,Jyy;
+
+   ASSERT_NE(sensorModel, nullptr);
    sensorModel->distortionJacobian(imagePt.samp, imagePt.line, Jxx, Jxy,Jyx,Jyy);
```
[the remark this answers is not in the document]
twilson271828: Fixed.

tests/FrameCameraTests.cpp
twilson271828: Fixed.

tests/FrameCameraTests.cpp
twilson271828: Fixed.

tests/FrameCameraTests.cpp
twilson271828: Fixed.

tests/FrameCameraTests.cpp
twilson271828: Fixed.

tests/FrameCameraTests.cpp
```diff
@@ -13,6 +13,61 @@
 using namespace std;
 using json = nlohmann::json;
 
+
+class FramerParameterizedTest : public ::testing::TestWithParam<csm::ImageCoord> {
```
jessemapel: I think we should parameterize based on the keyword in the ISD being modified instead of the image point.

tests/FrameCameraTests.cpp
```diff
@@ -94,6 +163,57 @@ class FrameSensorModel : public ::testing::Test {
       }
 };
 
+
+
+INSTANTIATE_TEST_CASE_P(JacobianTest,FramerParameterizedTest,
```
jessemapel: I think we want to change the first argument for this to something more general. Right now the test name will be
`JacobianTest/FramerParameterizedTest.JacobianTest/#`. Changing it to something like `FramingSensorModel` would give us a test called `FramingSensorModel/FramerParameterizedTest.JacobianTest/#`.

tests/FrameCameraTests.cpp
twilson271828: That would help clean up many of Kaj's tests, but I can think of plenty of situations where we want to parameterize on the values a keyword can take (i.e. Checking some distortion function with different values for a focal length, or passing in a range of double vectors for the keywords odt_x and odt_y).  My goal here was to put in something that is as easy to follow for others as possible. Parameterizing on an image pt was about as simple as I could make it.

jessemapel: :+1: that sounds reasonable

jessemapel: Agreed. I'm imagining we select 3-5 test points and run image to ground/ ground to image on each of those tests points and parameterize by the ISD keyword.

So, we'd have tests called "CenterPixel", "OffCenter", "OffBody", etc.

Then we would parameterize using pairs of ISD keywords and values.

jessemapel: I think this decision is going to be up to whomever does this next FY

twilson271828: That sounds good.

PR #117, opened 2018-10-03
UsgsAstroFramePlugin refactor, part one

### LINKED ISSUE #110: Update the framer plugin to use the new ISD format instead of the old Socet keywords format

kberryUSGS:
This includes the updates to `UsgsAstroFramePlugin` to switch from expecting the SET to parse a json ISD into a `csm::Isd` object that will be passed to this plugin to expecting the SET to construct a `csm::Isd('image file name')` to pass to the plugin. Now the plugin parses the json isd. It looks for a json isd file named 'filename.json' next to the provided image filename. This should fix #110 

The existing tests needed to be updated for this change. 

Now, `canModelBeConstructedFromISD` works on both 'filename only' `csm::Isd` objects and ones that have already been populated with values from the json isd file. 

I added a helper function I called `csm::Isd loadImageSupportData(const csm::Isd&) const` which takes a 'filename only' `csm::Isd` and creates and returns a new Isd populated with values from the associated json isd file. Is this function-name confusing, or all-right as is?

### DIFF
--- a/src/UsgsAstroFramePlugin.cpp
+++ b/src/UsgsAstroFramePlugin.cpp
@@ -278,6 +278,9 @@ return sensor_model;
 }
 
 
+// This function takes a csm::Isd which only has the image filename set. It uses this filename to
+// find a metadata json file loacated alongside the image file. It creates and returns new csm::Isd
+// with its parameters populated by the metadata file. 
 csm::Isd UsgsAstroFramePlugin::loadImageSupportData(const csm::Isd &imageSupportDataOriginal) const{
   // Get image location from the input csm::Isd: 
   std::string imageFilename = imageSupportDataOriginal.filename(); 
@@ -290,22 +293,29 @@ csm::Isd UsgsAstroFramePlugin::loadImageSupportData(const csm::Isd &imageSupport
   csm::Isd imageSupportData(isdFilename);
   imageSupportData.clearAllParams();
 
-  std::ifstream isdFile(isdFilename); 
-  json jsonIsd = json::parse(isdFile);
-   for (json::iterator it = jsonIsd.begin(); it != jsonIsd.end(); ++it) {
-    if (it.value().size() >1 ) {
-      std::vector<double> v = it.value();
-      for (int j=0;j < v.size(); j++) {
-        std::ostringstream val;
-        val << std::setprecision(15) << v[j];
-        imageSupportData.addParam(it.key(),val.str());
+  try {
+    std::ifstream isdFile(isdFilename); 
+    json jsonIsd = json::parse(isdFile);
+     for (json::iterator it = jsonIsd.begin(); it != jsonIsd.end(); ++it) {
+      if (it.value().size() >1 ) {
+        std::vector<double> v = it.value();
+        for (int j=0;j < v.size(); j++) {
+          std::ostringstream val;
+          val << std::setprecision(15) << v[j];
+          imageSupportData.addParam(it.key(),val.str());
+        }
+      }
+      else {
+        imageSupportData.addParam(it.key(), it.value().dump());
       }
     }
-    else {
-      imageSupportData.addParam(it.key(), it.value().dump());
-    }
+    isdFile.close(); 
+  } catch (...) {
+    std::string errorMessage = "Could not read metadata file associated with image: ";
+    errorMessage.append(isdFilename);
+    throw csm::Error(csm::Error::FILE_READ, errorMessage, 
+                     "UsgsAstroFramePlugin::loadImageSupportData"); 
   }
-  isdFile.close(); 
 
   return imageSupportData; 
 }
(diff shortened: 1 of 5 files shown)


jessemapel:
You'll need to update the `canModelBeConstructedFromISD` method so that also does the bit of fancy loading that you do at the start of `constructModelFromISD` right now, it is just getting an ISD without any parameters and is saying it's unusable because it doesn't know how to load the actual metadata from that.

kberryUSGS:
@jessemapel Thanks for catching this! I thought I'd moved `canModelBeConstructedFromIsd` everywhere it was used to _after_ the isd had the parameters loaded, but not in this test!

jessemapel:
@kberryUSGS It's expected that SETs will call that function with the "filename only" ISD, so it needs to be able to handle that.

kberryUSGS:
@jessemapel Now this is actually ready for review. I updated the initial blurb at the top of the PR with more up-to-date information.

src/UsgsAstroFramePlugin.cpp
```diff
@@ -273,18 +278,55 @@ return sensor_model;
...
+  
+  // Load 'sidecar' ISD file
+  size_t lastIndex = imageFilename.find_last_of("."); 
+  std::string baseName = imageFilename.substr(0, lastIndex); 
+  std::string isdFilename = baseName.append(".json");
+
+  csm::Isd imageSupportData(isdFilename);
+  imageSupportData.clearAllParams();
+
+  std::ifstream isdFile(isdFilename);
```
jessemapel: We should put a try/catch around this file IO in case the file doesn't exist or there's an issue with it. I don't think the ifstream error is going to be descriptive enough.

src/UsgsAstroFramePlugin.cpp
```diff
@@ -537,9 +578,14 @@ bool UsgsAstroFramePlugin::canISDBeConvertedToModelState(const csm::Isd &imageSu
       convertible = false;
   }
 
+  csm::Isd localImageSupportData = imageSupportData; 
+  if (imageSupportData.parameters().size() <= 0) {
```
jessemapel: This check should use multimap's empty() method instead of checking the size.

src/UsgsAstroFramePlugin.cpp
```diff
@@ -273,18 +278,55 @@ return sensor_model;
 }
 
 
-csm::Model *UsgsAstroFramePlugin::constructModelFromISD(const csm::Isd &imageSupportData,
+csm::Isd UsgsAstroFramePlugin::loadImageSupportData(const csm::Isd &imageSupportDataOriginal) const{
```
jessemapel: The name for this method (and its signature) is a little bit confusing, but I can't think of anything better for it. A few lines documenting what this does and why it's needed would be helpful.

tests/FrameCameraTests.cpp
```diff
@@ -39,118 +38,56 @@ class FramerParameterizedTest : public ::testing::TestWithParam<csm::ImageCoord>
...
-               isd.addParam(it.key(), jsonValue[i].dump());
-            }
-         }
-         else {
-            isd.addParam(it.key(), jsonValue.dump());
-         }
-      }
-      isdFile.close();
+    virtual void SetUp() {
+      isd.setFilename("data/simpleFramerISD.json");
```
jessemapel: It's expected that the filename will be the image filename. It's not a functional difference, but I think it's a better test of expected functionality if this ends in `.img` or `.cub` or `.raw`

tests/FrameCameraTests.cpp
```diff
@@ -39,118 +38,56 @@ class FramerParameterizedTest : public ::testing::TestWithParam<csm::ImageCoord>
...
-               }
-            }
-            else {
-               isd.addParam(it.key(), jsonValue.dump());
-            }
-         }
-
-         isdFile.close();
 
+         isd.setFilename("data/simpleFramerISD.json");
```
jessemapel: See comments on the other filename

tests/FrameCameraTests.cpp
```diff
@@ -159,64 +96,39 @@ class FrameSensorModel : public ::testing::Test {
...
-      isd.addParam("odt_y", strval.str());
-   }
-
    UsgsAstroFrameSensorModel* sensorModel = createModel(isd);
+   std::string modelState = sensorModel->getModelState(); 
+   auto state = json::parse(modelState);
 
+   state["m_odtX"] = {1.0, 0.0, 1.0, 0.0, 1.0, 0.0, 1.0, 0.0, 1.0, 0.0};
+   state["m_odtY"] = {0.0, 1.0, 0.0, 1.0, 0.0, 1.0, 0.0, 1.0, 0.0, 1.0};
+   sensorModel->replaceModelState(state.dump());
```
jessemapel: :+1: wooo portable tests

tests/FrameCameraTests.cpp
kberryUSGS: :+1:

tests/FrameCameraTests.cpp
kberryUSGS: :+1:

src/UsgsAstroFramePlugin.cpp
kberryUSGS: :+1:

kberryUSGS:
@jessemapel I updated this. One thing is that I did put in that I'm not sure about, especially after talking to Trent in the meeting this morning, is an error throw in the ifstream catch -- I could instead put in a csm::Warning::DATA_NOT_AVAILABLE.

Issue #120, opened 2018-10-03
Generate Line Scan tests


jessemapel:
Similar to how we tested the framing camera, we need generic line scan tests.

jessemapel:
Current testing plan:

- A linear flyby with a fixed viewing angle
- A fixed position with a constant camera rotation velocity

I'm going to try and generate some test numbers for this.

jessemapel:
[csm_ls_test.pdf](https://github.com/USGS-Astrogeology/CSM-CameraModel/files/2443342/csm_ls_test.pdf)

Here's the LS test case math. I feel good about all of it, except maybe the final rotational velocity for the second case. I think that's how rotational velocity works?

jlaura:
Linescan tests:

- [ ] i2g / g2i 
- [ ] off body
- [ ] A linear flyby with a fixed viewing angle
- [ ] A fixed position with a constant camera rotation velocity
- [ ] State test / resetting the state - dump state;reset;fail
- [ ] Test distortion (tests #105)
- [ ] Tests for #165
- [ ] Tests for #166
- [ ] Test imageToRemoteImagingLocus
- [ ] Test imageToProximateImagingLocus
- [ ]

jessemapel:
Example math for the [constant velocity](https://raw.githubusercontent.com/wiki/USGS-Astrogeology/usgscsm/documents/LS_Image_To_Ground_Example.pdf) and [constant angular velocity](https://raw.githubusercontent.com/wiki/USGS-Astrogeology/usgscsm/documents/pan_ls_math.pdf)

PR #132, opened 2018-10-12
First pass at LS tests

jessemapel:
This has the ISD and test for the constant velocity line scan test case in it. Right now, they are all failing because of a bug in how the line scan camera handles time, it's off by half a pixel.

I'm putting this up so it's here, but it should not be merged as it's going to fail CI.


jessemapel:
This is part of #120

PR #134, opened 2018-10-12
Fixed canModelBeConstructedFromISD throwing an exception when the metadata file doesn't exist.

### LINKED ISSUE #133: GXP returns a error for all images it loads
I think the Framer/Scanner needs to not return an error. Perhaps not even a warning. But the application does need to know it didn't work. I thinking looking at the original code might list what we need to return.

here is the error
 Line 1336 : DEBUG : ERROR: Could not read metadata file associated with image: C:\Program Files\BAE SYSTEMS\SOCET GXP 4.3.0\Config\VCT\Maps\TrueMarble\TrueMarble-2km-21600x10800.json-UsgsAstroFramePlugin::loadImageSupportData

The good news is that GXP loaded the DLL!

jessemapel:
This should fix #133 it will need to be built on windows to test for sure first.

### DIFF
--- a/src/UsgsAstroFramePlugin.cpp
+++ b/src/UsgsAstroFramePlugin.cpp
@@ -7,7 +7,7 @@
 #include <iostream>
 #include <sstream>
 #include <fstream>
-#include <map> 
+#include <map>
 
 #include <math.h>
 #include <csm.h>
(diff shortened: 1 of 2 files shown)


jessemapel:
We should also add a check to try and construct something from a non-metadata json file, but that's not what #133 is running into.

PR #135, opened 2018-10-12
Adds support for a windows build via appveyor

jlaura:


jlaura:
I need to get a pr in to the csm before this is going to work on appveyor. I am also not 100% sure that the #defines are truly needed. I am not going to have time to test that right now and will open a question issue.

.appveyor.yml
```diff
@@ -7,18 +7,11 @@ platform:
...
-  # - cmd: conda install -y -c usgs-astrogeology libcsm
-  # - cmake .
+  - git submodule update --init --recursive
+
 build_script:
-  # - cmake --build . -- %MSBUILD_ARGS%
-  - echo pass
+  - cmake -G "Visual Studio 15 2017 Win64" -DBUILD_TESTS=OFF ..
+  - cmake --build . --target ALL_BUILD --config Release
+
```
jessemapel: Is there a reason there's no on success conda build here?

jlaura: yes. I want to make sure that the build is working and then I need to figure out what / how to push this. We need the dll to go for GXP, but we need the dll and lib to go otherwise.

CMakeLists.txt
```diff
@@ -18,6 +18,9 @@ else()
 			    PATH_SUFFIXES "csm" 
 			    PATHS $ENV{CONDA_PREFIX}/include/)
   find_library(CSM_LIBRARY csmapi PATHS $ENV{CONDA_PREFIX}/lib)
+  
+  message("--Found CSM Library: ${CSM_LIBRARY}")
+  message("--Found CSM Include Directory: ${CSM_INCLUDE_DIR}")
```
jlaura: This is just cleanup on formatting.

CMakeLists.txt
```diff
@@ -42,24 +45,18 @@ target_include_directories(usgscsm
...
                       ${CSM_LIBRARY}
                       gtest ${CMAKE_THREAD_LIBS_INIT})
 
 if(WIN32)
-  option(CMAKE_USE_WIN32_THREADS_INIT "using WIN32 threads" ON)
-  option(gtest_disable_pthreads "Disable uses of pthreads in gtest." ON)
-  install(TARGETS usgscsm
-    RUNTIME DESTINATION ${CMAKE_INSTALL_LIBDIR})
-    install(DIRECTORY ${USGSCSM_INCLUDE_DIRS} DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
+  install(TARGETS usgscsm RUNTIME DESTINATION ${CMAKE_INSTALL_LIBDIR})
```
jlaura: This should remove the duplication. I am not 100% sure why RUNTIME is needed on win. I found this in a solution, so we should not consider this to be a best practice.

src/UsgsAstroLsPlugin.cpp
```diff
@@ -17,10 +17,10 @@
 //
 //-----------------------------------------------------------------------------
 
-#define USGSASTROLINESCANNER_LIBRARY
+#define USGS_SENSOR_LIBRARY
 
 #include "UsgsAstroLsPlugin.h"
-#include "UsgsAstroLsISD.h"
+//#include "UsgsAstroLsISD.h"
```
jlaura: This can get pulled officially in a cleanup PR. This will not compile on win without this.

tests/CMakeLists.txt
```diff
@@ -2,6 +2,12 @@ cmake_minimum_required(VERSION 3.10)
 
 # Link runCSMCameraModelTests with what we want to test and the GTest and pthread library
 add_executable(runCSMCameraModelTests FramePluginTests.cpp FrameCameraTests.cpp)
-target_link_libraries(runCSMCameraModelTests usgscsm ${GTEST_LIBRARIES} ${GTEST_MAIN_LIBRARIES} pthread)
+if(WIN32)
+  option(CMAKE_USE_WIN32_THREADS_INIT "using WIN32 threads" ON)
+  option(gtest_disable_pthreads "Disable uses of pthreads in gtest." ON)
```
jlaura: I moved these options from the main CMakeLists.txt into here since they are gtest specific.

jlaura: I guess that this is also moot since the tests are turned off.

jlaura:
Take2 on the rebase - the build is at least passing...

PR #138, opened 2018-10-12
Refactoring to move to one plugin and removed unused classes. 

Kelvinrr:
Tests still failing, need to clean up error checking code and make sure the pipelines are clean.


src/UsgsAstroPlugin.cpp
```diff
@@ -0,0 +1,306 @@
...
+      }
+  // Check that the necessary keys are there (this does not chek values at all.)
+  // This sort of check should probably be moved into the sensor model
+  // auto state = json::parse(modelState);
+  // for(auto &key : MODEL_KEYWORDS[model_name_from_state]){
+  //     if (state.find(key) == state.end()){
+  //         constructible = false;
+  //         break;
+  //     }
+  // }
```
jessemapel: I'm worried that if we don't check this here, some SETs will fail. They probably expect a false from this call and then will not call constructFromState. If they get a true from this, and then an exception from constructFromState they will probably completely fail.

jessemapel: Right now, the sensor model checks that the state has all of the required values in the replaceModelState function. I would move all of that into another function that would return a list of missing/invalid keywords. Then, in replaceModelState call that function to check the state is good. The plugin, can then call that same function here.

src/UsgsAstroPlugin.cpp
```diff
@@ -0,0 +1,306 @@
...
+bool UsgsAstroPlugin::canModelBeConstructedFromISD(const csm::Isd &imageSupportData,
+                                              const std::string &modelName,
+                                              csm::WarningList *warnings) const {
+  return canISDBeConvertedToModelState(imageSupportData, modelName, warnings);
+}
+
+
+// This function takes a csm::Isd which only has the image filename set. It uses this filename to
+// find a metadata json file loacated alongside the image file. It creates and returns new csm::Isd
+// with its parameters populated by the metadata file.
```
jessemapel: comment string is incorrect

src/UsgsAstroPlugin.cpp
```diff
@@ -0,0 +1,306 @@
...
+      throw(csmErr);
+  }
+
+  // if (MODEL_KEYWORDS.find(name) == MODEL_KEYWORDS.end()){
+  //     csm::Error::ErrorType aErrorType = csm::Error::SENSOR_MODEL_NOT_SUPPORTED;
+  //     std::string aMessage = "Sensor model not supported.";
+  //     std::string aFunction = "UsgsAstroPlugin::getModelNameFromModelState()";
+  //     csm::Error csmErr(aErrorType, aMessage, aFunction);
+  //     throw(csmErr);
+  // }
```
jessemapel: Why's this here? Do we need differentiate between invalid state for supported model and unsupported model?

src/UsgsAstroPlugin.cpp
```diff
@@ -0,0 +1,306 @@
...
+                                               csm::WarningList *warnings) const {
+  try {
+       convertISDToModelState(imageSupportData, modelName);
+  }
+  catch(...) {
+      // key didn't exist
+      return false;
+  }
+  return true;
+}
```
jessemapel: Ideally this function will return warnings about what's wrong when it returns false.

src/UsgsAstroPlugin.cpp
```diff
@@ -0,0 +1,306 @@
...
+
+  csm::Model* sensor_model = constructModelFromISD(
+                             imageSupportData, modelName);
+
+  if (sensor_model == 0) {
+      csm::Error::ErrorType aErrorType = csm::Error::ISD_NOT_SUPPORTED;
+      std::string aMessage = "ISD not supported: ";
+      std::string aFunction = "UsgsAstroPlugin::convertISDToModelState()";
+      throw csm::Error(aErrorType, aMessage, aFunction);
+  }
```
jessemapel: Chances are if it's not supported, constructModelFromISD is going to throw an exception. I don't know a case in which it would return NULL without an exception.

src/UsgsAstroPlugin.cpp
```diff
@@ -0,0 +1,306 @@
...
+
+
+std::string UsgsAstroPlugin::convertISDToModelState(const csm::Isd &imageSupportData,
+                                               const std::string &modelName,
+                                               csm::WarningList *warnings) const {
+
+  csm::Model* sensor_model = constructModelFromISD(
+                             imageSupportData, modelName);
+
+  if (sensor_model == 0) {
```
jessemapel: Null pointers are falsey, you can just say `if (!sensor_model)`

src/UsgsAstroPlugin.cpp
Kelvinrr: Most of it was commented out while I was trying to figure out the best way to handle these errors. The original was mostly spaghetti code and leaving them as they were caused stack overflow errors, but at the same time wasn't sure what exactly the CSM spec wanted in terms of where errors should be handled and how those errors should look like. This is pretty much the story behind most of the commented out errors and validation checking.

src/UsgsAstroPlugin.cpp
Kelvinrr: I know this, but doing that originally gave me a type error (expected bool, got Model*, as the error said) and so I just changed it rapidly not having the time and too tired to figure out why at the time. Probably something quirky was happening.

src/UsgsAstroPlugin.cpp
Kelvinrr: The spec seems to want that `csm::WarningList` pointer to be passed up and down the call stack being appended to. So from the insertion point for the SET, the pointer is being modified as it goes instead of being returned.

Issue #104, opened 2018-08-27
Refactor Framer tests to use new parameterized test setup


jessemapel:
We should refactor the tests Kaj wrote to use parameterized tests to condense our testing code. See #101 for a template of how we can do this.

PR #140, opened 2018-10-15
Minor refactor of tests that modify a state value

kberryUSGS:
This isn't a full refactor to use parameterized tests. Instead, for now, I removed a bunch of repeated code in tests that tweak a single state parameter's value and moved this to a function in the associated fixture. This is related to #104 in that it could be a first step in that direction, but doesn't fix it completely.


Kelvinrr:
Pulled locally and testsed, looks like you're missing a class name? FrameIsdTest->FrameStateTest in on line 73 in FramePluginTests.cpp

kberryUSGS:
@Kelvinrr Fixed!

thareUSGS:
👍

kberryUSGS:
@Kelvinrr Thanks :+1:

PR #146, opened 2018-10-19
Adds if/else into appveyor

jlaura:


.appveyor.yml
```diff
@@ -31,14 +31,16 @@ on_success:
...
         $anaconda_label = "main"
       };
-      Write-Host "anaconda_label $anaconda_label";
-      $parameters = '-t', "$env:CONDA_UPLOAD_TOKEN", 'upload', "$tar_glob", '-l',
-                    "$anaconda_label", '--force', '--no-progress';
-      & cmd /c 'anaconda 2>&1' $parameters;
+    Write-Host "anaconda_label $anaconda_label";
+    $parameters = '-t', "$env:CONDA_UPLOAD_TOKEN", 'upload', "$tar_glob", '-l',
+                  "$anaconda_label", '--force', '--no-progress';
+    & cmd /c 'anaconda 2>&1' $parameters;
```
jessemapel: These should be indented. I don't think yaml cares, but it's more readable.

jlaura: The `& cmd...` is a new command. So in the quoted, we have 3 independent commands. I think keep as is?

jlaura: Or do you mean that whole block - I think that is what you mean.

Issue #119, opened 2018-10-03
Refactor line scan camera to work with the refactored plugin


jessemapel:
After #118, we need to make the camera model compatible with the updated plugin.

PR #149, opened 2018-10-23
Changed line scan sensor model to new ISD spec

jessemapel:
This PR updates the conversion code to use the new spec. Because we still don't have full test coverage on the line scan model, I can't confirm that everything is working correctly in the sensor model. I've tried to check everything, but once we refactor the line scan model we'll need to very carefully check all this.

The new LS ISD also has some changes to the spec:
- The sensor_location, sensor_velocity, and sensor_orientation objects now have arrays in them.
- I added an integration time keyword
- I added a t0_ephemeris keyword

Unit conversion are not supported yet, but that can be copied and pasted from the framer.


src/UsgsAstroLsSensorModel.cpp
```diff
@@ -2664,57 +2660,56 @@ std::string UsgsAstroLsSensorModel::constructStateFromIsd(const std::string imag
...
+   state["m_totalLines"] = isd.at("image_lines");
+   state["m_totalSamples"] = isd.at("image_samples");
    state["m_offsetLines"] = 0.0;
    state["m_offsetSamples"] = 0.0;
-   state["m_platformFlag"] = isd.at("PLATFORM");
-   state["m_aberrFlag"] = isd.at("ABERR");
-   state["m_atmRefFlag"] = isd.at("ATMREF");
+   state["m_platformFlag"] = 1;
+   state["m_aberrFlag"] = 0;
+   state["m_atmRefFlag"] = 0;
```
jlaura: I had a note about whether we should turn off or support. So :+1: on turning these off.

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -2664,57 +2660,56 @@ std::string UsgsAstroLsSensorModel::constructStateFromIsd(const std::string imag
...
-     int numIntTimes = isd.at("NUMBER_OF_INT_TIMES");
+     int numIntTimes = isd.at("number_of_int_times");
      for (int i = 0; i < numIntTimes; i++) {
-       state["m_intTimeLines"].push_back(isd.at("INT_TIME").at(i*3));
-       state["m_intTimeStartTimes"].push_back(isd.at("INT_TIME").at(i*3+1));
-       state["m_intTimes"].push_back(isd.at("INT_TIME").at(i*3+2));
+       // TODO what does this look like under the new spec?
+       state["m_intTimeLines"].push_back(isd.at("int_time").at(i*3));
+       state["m_intTimeStartTimes"].push_back(isd.at("int_time").at(i*3+1));
+       state["m_intTimes"].push_back(isd.at("int_time").at(i*3+2));
```
jlaura: @Kelvinrr should weigh in here on the specification: https://github.com/USGS-Astrogeology/pfeffernusse/blob/master/swagger.yaml

I *think* this is all inside of the `line_scan_rate` entry in the spec? This comment is for lines 2676 through 2685.

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -2664,57 +2660,56 @@ std::string UsgsAstroLsSensorModel::constructStateFromIsd(const std::string imag
...
      }
    }
 
-   state["m_centerEphemerisTime"] = isd.at("CENTER_EPHEMERIS_TIME");
-   state["m_detectorSampleSumming"] = isd.at("DETECTOR_SAMPLE_SUMMING");
-   state["m_startingSample"] = isd.at("STARTING_SAMPLE");
-   state["m_ikCode"] = isd.at("IKCODE");
-   state["m_focal"] = isd.at("FOCAL");
-   state["m_isisZDirection"] = isd.at("ISIS_Z_DIRECTION");
+   state["m_startingEphemerisTime"] = state["m_intTimes"].at(0);
```
jlaura: Right now we pass an explicit `starting_ephemeris_time` - I think we could refactor that out of the spec. We need center time for framers and the `line_scan_rate` should cover push* sensors.

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -2664,57 +2660,56 @@ std::string UsgsAstroLsSensorModel::constructStateFromIsd(const std::string imag
...
+   state["m_detectorSampleSumming"] = 1;
```
jlaura: Should come through as `detector_sample_summing`

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -2664,57 +2660,56 @@ std::string UsgsAstroLsSensorModel::constructStateFromIsd(const std::string imag
...
+   state["m_startingSample"] = 0;
```
jlaura: Should come through as `starting_detector_sample`

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -2664,57 +2660,56 @@ std::string UsgsAstroLsSensorModel::constructStateFromIsd(const std::string imag
...
+   state["m_ikCode"] = 0;
```
jlaura: I have not scrolled through, but doesn't this break LROC NAC? - ignore if the logic block below is updated.

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -2664,57 +2660,56 @@ std::string UsgsAstroLsSensorModel::constructStateFromIsd(const std::string imag
...
+   state["m_focal"] = isd.at("focal_length_model").at("focal_length");
+   state["m_isisZDirection"] = 1;
```
jlaura: Completely remove or incremental refactor so used in ecef call - ignore if updated below.

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -2664,57 +2660,56 @@ std::string UsgsAstroLsSensorModel::constructStateFromIsd(const std::string imag
...
 
    for (int i = 0; i < 3; i++)
    {
-      state["m_opticalDistCoef"][i] = isd.at("OPTICAL_DIST_COEF").at(i);
-      state["m_iTransS"][i] = isd.at("ITRANSS").at(i);
-      state["m_iTransL"][i] = isd.at("ITRANSL").at(i);
+      state["m_opticalDistCoef"][i] = isd.at("optical_distortion").at("r").at(i);
```
jlaura: `r` or `k` to better match what some of the kernels have been doing? (ODK).  Doesn't matter to me.

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -2664,57 +2660,56 @@ std::string UsgsAstroLsSensorModel::constructStateFromIsd(const std::string imag
...
-   double cos_b = cos(isd.at("MOUNTING_ANGLES").at(1).get<double>());
-   double sin_b = sin(isd.at("MOUNTING_ANGLES").at(1).get<double>());
-   double cos_c = cos(isd.at("MOUNTING_ANGLES").at(2).get<double>());
-   double sin_c = sin(isd.at("MOUNTING_ANGLES").at(2).get<double>());
+   double cos_a = cos(0);
+   double sin_a = sin(0);
+   double cos_b = cos(0);
+   double sin_b = sin(0);
+   double cos_c = cos(0);
+   double sin_c = sin(0);
```
jlaura: :+1:

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -2727,60 +2722,51 @@ std::string UsgsAstroLsSensorModel::constructStateFromIsd(const std::string imag
...
-       state["m_ephemRates"].push_back(isd.at("EPHEM_RATES").at(i));
-   }
-
-   int numQuat = isd.at("NUMBER_OF_QUATERNIONS");
-   for (int i=0; i < numQuat * 4; i++){
-       state["m_quaternions"].push_back(isd.at("QUATERNIONS").at(i));
+   state["m_dtEphem"] = isd.at("dt_ephemeris");
+   state["m_t0Ephem"] = isd.at("t0_ephemeris");
+   state["m_dtQuat"] =  isd.at("dt_ephemeris");
+   state["m_t0Quat"] =  isd.at("t0_ephemeris");
```
jlaura: quaternion? - line above too.

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -2727,60 +2722,51 @@ std::string UsgsAstroLsSensorModel::constructStateFromIsd(const std::string imag
...
+
+   state["m_numEphem"] = isd.at("number_of_ephemerides");
+   state["m_numQuaternions"] = isd.at("number_of_ephemerides");
```
jlaura: quaternion?

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -2727,60 +2722,51 @@ std::string UsgsAstroLsSensorModel::constructStateFromIsd(const std::string imag
...
+       state["m_ephemPts"].push_back(isd.at("sensor_location").at("x").at(i));
+       state["m_ephemPts"].push_back(isd.at("sensor_location").at("y").at(i));
+       state["m_ephemPts"].push_back(isd.at("sensor_location").at("z").at(i));
+       state["m_ephemRates"].push_back(isd.at("sensor_velocity").at("x").at(i));
+       state["m_ephemRates"].push_back(isd.at("sensor_velocity").at("y").at(i));
+       state["m_ephemRates"].push_back(isd.at("sensor_velocity").at("z").at(i));
+       state["m_quaternions"].push_back(isd.at("sensor_orientation").at("w").at(i));
+       state["m_quaternions"].push_back(isd.at("sensor_orientation").at("v1").at(i));
+       state["m_quaternions"].push_back(isd.at("sensor_orientation").at("v2").at(i));
+       state["m_quaternions"].push_back(isd.at("sensor_orientation").at("v3").at(i));
```
jlaura: A bigger question - when might one not have number_of_ephemerides and number_of_quaternions not in sync? If the answer is never, we can drop number_of_quaternions from the spec.

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -2727,60 +2722,51 @@ std::string UsgsAstroLsSensorModel::constructStateFromIsd(const std::string imag
...
+   state["m_trajectoryIdentifier"] = "UNKNOWN";
+   state["m_collectionIdentifier"] = "UNKNOWN";
 
    // Ground elevations
-   state["m_refElevation"] = isd.at("REFERENCE_HEIGHT");
-   state["m_minElevation"] = isd.at("MIN_VALID_HT");
-   state["m_maxElevation"] = isd.at("MAX_VALID_HT");
+   state["m_refElevation"] = 0.0;
+   state["m_minElevation"] = isd.at("reference_height").at("minheight");
+   state["m_maxElevation"] = isd.at("reference_height").at("maxheight");
```
jlaura: I have a question on the pfeffernusse side about what to send out here. For now, I am defaulting to meaningless floats (0,100) I think.

src/UsgsAstroLsSensorModel.cpp
jessemapel: This part will get removed eventually so for right now just sets all the mounting angles to 0

src/UsgsAstroLsSensorModel.cpp
jessemapel: It depends on if we want to have position and pointing at different times. I could see requiring position and pointing to be at the same times.

If we go with possibly separate times then you're correct about using dt_quaternion and t0_quaternion

src/UsgsAstroLsSensorModel.cpp
jessemapel: See other comment

jessemapel:
Fixes #119

src/UsgsAstroLsSensorModel.cpp
Kelvinrr: If these values don't matter, do we still need to set them at all?

jessemapel: The aberration correction and atmosphere flags can be removed. The platform identifier is actually part of the CSM spec.

You're right the next step after this PR is to prune out a lot of these un-needed values.

tests/data/constVelocityLineScan.json
```diff
@@ -1,111 +1,87 @@
...
+      [0.0, 0.0, -20.0],
+      [0.0, 0.0, -20.0],
+      [0.0, 0.0, -20.0],
+      [0.0, 0.0, -20.0],
+      [0.0, 0.0, -20.0]
+    ]
+  },
+  "starting_detector_line": 0,
+  "starting_detector_sample": 0,
+  "sun_position": [100.0, 100.0, 0.0],
```
jlaura: Should `sun_position` and `sun_velocity` use the same objects are `sensor_velocity`. That should help with unit management, etc.

jessemapel: :+1:

src/UsgsAstroFrameSensorModel.cpp
```diff
@@ -821,71 +824,68 @@ std::string UsgsAstroFrameSensorModel::constructStateFromIsd(const std::string&
 
         state["m_focalLength"] = focal_length;
         state["m_focalLengthEpsilon"] = epsilon;
```
jlaura: `focal_epsilon`?

src/UsgsAstroFrameSensorModel.cpp
```diff
@@ -821,71 +824,68 @@ std::string UsgsAstroFrameSensorModel::constructStateFromIsd(const std::string&
...
 
       // get sun_position
+      // sun position is not strictly necessary, but is required for getIlluminationDirection.
       {
         json jayson = isd.at("sun_position");
-        json x = jayson.at("x");
-        json y = jayson.at("y");
-        json z = jayson.at("z");
+        json positions = jayson.at("positions")[0];
+        json unit = jayson.at("unit");
```
jlaura: Do we take in the sun velocities even though they are unused?

src/UsgsAstroFrameSensorModel.cpp
```diff
@@ -821,71 +824,68 @@ std::string UsgsAstroFrameSensorModel::constructStateFromIsd(const std::string&
...
-        state["m_currentParameterValue"][4] = isd.at("sensor_orientation").at(1);
-        state["m_currentParameterValue"][5] = isd.at("sensor_orientation").at(2);
-        state["m_currentParameterValue"][6] = isd.at("sensor_orientation").at(3);
+        json jayson = isd.at("sensor_orientation");
+        json quaternion = jayson.at("quaternions")[0];
+
+        state["m_currentParameterValue"][3] = quaternion[0];
+        state["m_currentParameterValue"][4] = quaternion[1];
+        state["m_currentParameterValue"][5] = quaternion[2];
+        state["m_currentParameterValue"][6] = quaternion[3];
```
jlaura: So this looks like an array of arrays in form `[[<all 0>], [<all 1>], [<all 2>], [<all 3>]]`. The specification is `[[q1], [q2], [q3], ..., [qn]]`. Is this setting the parameter value for just the first quaternion.

src/UsgsAstroFrameSensorModel.cpp
jessemapel: We should. Right now this code isn't setup well for optional things. I'll look into a clean way to do this.

src/UsgsAstroFrameSensorModel.cpp
jessemapel: Yes, line 867, pull off the first set of quaternions (q1) and then this parses those into the current parameters

src/UsgsAstroFrameSensorModel.cpp
jessemapel: Because they are currently unused, I think we should continue to ignore them. If we add support for them later as an optional value, then we should start bringing them in.

src/UsgsAstroFrameSensorModel.cpp
jessemapel: Yeah, I'll change that.

src/UsgsAstroFrameSensorModel.cpp
jlaura: :+1:

src/UsgsAstroFrameSensorModel.cpp
jlaura: :+1:

jlaura:
Once the `focal_epsilon` is updated feel free to merge. No reason to wait on one of us to push the big green button.

Issue #142, opened 2018-10-15
GXP not recognizing json ISD ...


thareUSGS:
I have two testing images for MDIS NAC and CTX with ISDs from pfeffernusse.

 /usgs/shareall/thare/SocetCode/GXP_CSM_build/CTX_test/
and
/usgs/shareall/thare/SocetCode/GXP_CSM_build/MDIS_test/

I converted to Tiff and ISIS cubes for GXP testing.

In the log, during "debug" startup, I get the CSM success for the latest usgscam.dll
_Succeeded to load CSM plugin library: C:\Program Files\BAE SYSTEMS\SOCET GXP 4.3.0\bin\smplugins\csmplugins\3.0.3\usgscsm.dll_

Image are loaded fine and the auto pyramids and histogram files are run, but the images are set to no camera model. 

I don't see anything specific to our code in the logs besides a generic:
_File SenFSensorModelFactory.cpp 	Line 646 : DEBUG : createSensorModel( ... ) - Unable to create CSM or CSM30 Sensor Model_

which implies it ran through all the CSM dlls and failed on all of them. GXP doesn't report failure per CSM. I do know if a CSM is found, it will report that as a success in the log. For example, with the original CSM, it will report:

_SenSocCsm30SensorModel.cpp 	Line 1361 : INFO : Available CSM: USGS_ASTRO_LINE_SCANNER_SENSOR_MODEL for image file: ..._

I wonder if someone can try latest code on these images and available ISD json files.

thareUSGS:
GXP still isn't happy with the latest build. There is nothing obvious in the logs failing. The image loads fine, but no sensor model is associated. Have we actually tried the code to use the "load the image" and find the *.json file to fire associate the CSM? I don't recall seeing this tested.

a 16x16 pixel tif and the generic linescan ISD are here:
/usgs/shareall/thare/SocetCode/GXP_CSM_build/JSON_CTX_test/

jessemapel:
@thareUSGS The plugin currently does that and all of the tests are confirming that it works.

thareUSGS:
With GXP debug mode on it appears I am getting the new messages. Looks like we might be missing a CSM function. I hope this isn't just a GXP thing. Here are the errors:

```
[...]

PR #151, opened 2018-10-31
Fixes 142 for framer

jlaura:


src/UsgsAstroPlugin.cpp
```diff
@@ -154,12 +154,15 @@ std::string UsgsAstroPlugin::loadImageSupportData(const csm::Isd &imageSupportDa
   std::string imageFilename = imageSupportDataOriginal.filename();
   size_t lastIndex = imageFilename.find_last_of(".");
   std::string baseName = imageFilename.substr(0, lastIndex);
+  lastIndex = baseName.find_last_of("/");
```
jessemapel: This part isn't going to be portable to windows. We'll have to use the dir_separator compiler flag bit

jlaura: :+1: Fixed in an inbound PR.

jlaura:
@Kelvinrr @jessemapel The PR is passing / working, but is having issues. I have to step out for a bit. I am not sure how to handle the `m_referencePoint` and `m_currentParameterCovariance` members in the `replaceModelState` and `getModelState` methods.

src/UsgsAstroFrameSensorModel.cpp
```diff
@@ -657,9 +651,13 @@ std::string UsgsAstroFrameSensorModel::getModelState() const {
...
                                    m_currentParameterValue[6]}},
-      {"m_currentParameterCovariance", m_currentParameterCovariance}
+      {"m_imageIdentifier", m_imageIdentifier},
+      {"m_collectionIdentifier", m_collectionIdentifier}
     };
-
+    state["m_referencePointXyz"] = json();
+    state["m_referencePointXyz"]["x"] = m_referencePointXyz.x;
+    state["m_referencePointXyz"]["y"] = m_referencePointXyz.y;
+    state["m_referencePointXyz"]["z"] = m_referencePointXyz.z;
```
Kelvinrr: In `contructStateFromIsd` this is being initialized as a `std::vector<float>`, probably best to stick with the vector since that's how most of these values are being handled and only translate to the point object when we have to, (i.e. in `getReferencePoint()`)

src/UsgsAstroFrameSensorModel.cpp
```diff
@@ -657,9 +651,13 @@ std::string UsgsAstroFrameSensorModel::getModelState() const {
                                    m_currentParameterValue[2], m_currentParameterValue[3],
                                    m_currentParameterValue[4], m_currentParameterValue[5],
                                    m_currentParameterValue[6]}},
-      {"m_currentParameterCovariance", m_currentParameterCovariance}
+      {"m_imageIdentifier", m_imageIdentifier},
+      {"m_collectionIdentifier", m_collectionIdentifier}
```
Kelvinrr: `m_currentParameterCovariance` no longer being added to state?

src/UsgsAstroFrameSensorModel.cpp
```diff
@@ -147,9 +152,17 @@ csm::ImageCoordCovar UsgsAstroFrameSensorModel::groundToImage(const csm::EcefCoo
...
+    gp.x = groundPt.x;
+    gp.y = groundPt.y;
+    gp.z = groundPt.z;
+
+    csm::ImageCoord ip = groundToImage(                                          
+      gp, desiredPrecision, achievedPrecision, warnings);                     
+   csm::ImageCoordCovar result(ip.line, ip.samp);                                    
+   // This is a partial, incorrect implementation to test if SocetGXP needs
+   // this method implemented in order to load the sensor.
+   return result;
```
jlaura: This is important that this is not a full implementation, just a test for GXP compliance.

src/UsgsAstroFrameSensorModel.cpp
```diff
@@ -210,9 +223,9 @@ csm::EcefCoordCovar UsgsAstroFrameSensorModel::imageToGround(const csm::ImageCoo
                                   double heightVariance, double desiredPrecision,
                                   double *achievedPrecision,
                                   csm::WarningList *warnings) const {
-    throw csm::Error(csm::Error::UNSUPPORTED_FUNCTION,
-      "Unsupported function",
-      "UsgsAstroFrameSensorModel::imageToGround");
+    // This is an incomplete implementation to see if SocetGXP needs this method implemented.
+    csm::EcefCoordCovar result;
+    return result;
```
jlaura: As above, basically testing to see if we need to return a result instead of throwing.

src/UsgsAstroFrameSensorModel.cpp
```diff
@@ -550,52 +562,38 @@ std::string UsgsAstroFrameSensorModel::getModelName() const {
 
 
 std::string UsgsAstroFrameSensorModel::getPedigree() const {
-  throw csm::Error(csm::Error::UNSUPPORTED_FUNCTION,
-                   "Unsupported function",
-                   "UsgsAstroFrameSensorModel::getPedigree");
+  return "USGS_FRAMER";
```
jlaura: Poor form, but again, just testing for GXP.

jlaura:
@all I put comments in about why some of the items are in as 'dummy' returns. This address #142 and #164 in that this should be able to be instantiated by GXP. The implementation though needs to be updated to provide the correct returns.

jessemapel:
I'm fine merging this, can you or @Kelvinrr generate issues for all of the methods that need to be implemented?

Kelvinrr:
I believe it'll just be `imageToGround`. The rest seem fine as they exist in this PR. Probably not worth creating a separate issue for now.

PR #156, opened 2018-11-29
Updated ISD schema to match as best I could to current pfeffer drivers

thareUSGS:
Updated ISD schema to match as best I could to current drivers in pfeffernusse.

See schema testing and validation tests in this Gist: https://gist.github.com/thareUSGS/ce936092699009da8f1f1e0247917013

### DIFF
--- a/schemas/isd.json
+++ b/schemas/isd.json
@@ -3,11 +3,51 @@
    "title": "Instrument Specification Definition",
    "description": "The Instrument Specification Definition for Astro CSM",
    "type": "object",
-
+   "required": [
+     "name_model",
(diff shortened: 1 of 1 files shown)


jessemapel:
No this isn't being used anywhere

thareUSGS:
hmm - so how do I validate an existing ISD without a schema definition? The swaggerhub link is not a schema definition? Does/can Pfeffernusse have a validation service for existing ISDs?

jlaura:
@thareUSGS The link about should be the schema, in read only form. Is it not for you?

PR #158, opened 2018-12-19
Merging dev into master for release

jlaura:


jlaura:
@jessemapel After squash and merge, this needs to have a release peeled off.  Since we are using the [Semantic Versioning](https://semver.org) we need to determine a version number for this initial release. I would be inclined to use a 0.1.0 release since we are still in a beta phase with our API.

jessemapel:
Because we are having external people using this, I think we're actually at 1.0 now

PR #159, opened 2018-12-19
Updates the version number in the meta.

jlaura:


recipe/meta.yaml
```diff
@@ -1,12 +1,13 @@
...
   name: usgscsm
-  version: "0.1.0"
+  version: {{ environ.get("GIT_DESCRIBE_TAG", "") }}
 
 source:
   git_url: https://github.com/USGS-Astrogeology/CSM-CameraModel.git
 build:
-  string: dev
+  number: {{ environ.get("GIT_DESCRIBE_NUMBER", 0) }}
+  string: {{ environ.get("GIT_DESCRIBE_HASH", "") }}
```
jessemapel: We don't need both the number of commits since tag and the current commit. We can just use the current HEAD commit.

jlaura:
@jessemapel This should be good to go now; I pulled the build number so just the hash on the head.

recipe/meta.yaml
```diff
@@ -1,12 +1,12 @@
 
 package:
   name: usgscsm
-  version: "0.1.0"
```
jessemapel: Is there a reason why we are replacing the version number with the commit hash too? I'd image we want new_version.commit_hash for our dev builds. So right now, our dev builds would be 1.1.0.#######

jlaura: The GIT_DESCRIBE_TAG should be '1.0.0' based on the 12.19.18 release. Then we should be seeing this getting pushed as `1.0.0-<GIT_DESCRIBE_HASH>-0` as far as I understand.  Our last tag is 1.0.0 and we are overriding the base conda string with the git env variable. The alternative would be to use the `tag-<default_conda_has>-<GIT_DESCRIBE_NUMBER>` to count the number of commits since the last tag. Both will be unique for the dev branch though the former probably provides better info if we are debugging an issue since we will have the short hard.

jessemapel: I think this is acceptable then.

Issue #105, opened 2018-08-28
Line Scan Optical Distortion


jlaura:
The line scan camera model does not implement optical distortion in computeViewingPixel (which is called by groundToImage). Therefore, when the optical distortion is passed in, we can run imageToGround without issue, but can not invert the process.  Setting optical distortion to [0,0,0] allows g2i to 'function', albeit incorrectly.

As a rider, it does not look like the optical distortion implemented in the code matches the model described in the IK. This is from a cursory look and could be entirely incorrect.

jessemapel:
I'm not sure if this is a bug or an enhancement because it is a known issue it just was never implemented?

jessemapel:
@jlaura The math does line up with the calculations from the CTX IK. It's the only distortion is can do right now, but that's what it has.

jessemapel:
@jlaura Do you have a CTX ISD somewhere that I can test this with?

PR #161, opened 2019-01-28
Inverts polynomial radial distortion in Line Scan ground to image

jessemapel:
This is largely a straight rip from ISIS3. The distortion for image to ground is also largely a rip from ISIS3.

I can make some better comments/documentation if needed.

### DIFF
--- a/src/UsgsAstroLsSensorModel.cpp
+++ b/src/UsgsAstroLsSensorModel.cpp
@@ -2452,10 +2452,47 @@ csm::ImageCoord UsgsAstroLsSensorModel::computeViewingPixel(
    double focalX = correctedLookX * lookScale;
    double focalY = correctedLookY * lookScale;
 
-   // TODO invert distortion here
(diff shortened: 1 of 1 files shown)


jessemapel:
Fixes #105

jessemapel:
For future reference, this method iteratively adds distortion until the output distorted pixel undistorts back to the input undistorted pixel.

jessemapel:
@thareUSGS The image to ground distortion code is already locked to only radial distortion :(

thareUSGS:
+1. Just a couple references (besides what is already available in ISIS3).
- from us: https://github.com/USGS-Astrogeology/usgscsm/blob/dev/jupyter_notebooks/UsgsAstroFrameSensorModelTesting.ipynb
- wikipedia showing my favorite Magnum PI mustache dist: https://en.wikipedia.org/wiki/Distortion_(optics)
- just interesting: http://www.ipol.im/pub/art/2011/ags-alde/

PR #162, opened 2019-01-29
Fixed ground to image and image to ground not giving inverse results.

jessemapel:
This also adds a very simple test to the line scan camera and some test stubs.

### DIFF
--- a/src/UsgsAstroLsSensorModel.cpp
+++ b/src/UsgsAstroLsSensorModel.cpp
@@ -525,6 +525,8 @@ csm::ImageCoord UsgsAstroLsSensorModel::groundToImage(
    // converge significantly faster, but it can be slower than simple bisection
    // in some situations.
 
+   std::cerr << std::setprecision(15) << "--== Starting Ground to Image! ==--" << std::endl;
+
    // Start bisection search on the image lines
    double sampCtr = m_totalSamples / 2.0;
    double firstTime = getImageTime(csm::ImageCoord(0.0, sampCtr));
@@ -561,11 +563,22 @@ csm::ImageCoord UsgsAstroLsSensorModel::groundToImage(
    // Increase the precision by a small amount to ensure the desired precision is met
    double pixelPrec = desired_precision / approxLineRes * 0.9;
 
+   std::cerr << std::setprecision(15)<< "Starting Bisection Search!" << std::endl;
+
    // Start bisection search for zero
    for (int it = 0; it < 30; it++) {
+      std::cerr << std::setprecision(15) << std::endl  << "Starting Iteration " << it+1 << std::endl;
       double nextTime = ((firstTime * lastOffset) - (lastTime * firstOffset))
                       / (lastOffset - firstOffset);
       double nextOffset = computeViewingPixel(nextTime, ground_pt, adj).line - 0.5;
+
+      std::cerr << std::setprecision(15) << "First time:" << firstTime << std::endl;
+      std::cerr << std::setprecision(15) << "Last time:" << lastTime << std::endl;
+      std::cerr << std::setprecision(15) << "Next time " << nextTime << std::endl;
+      std::cerr << std::setprecision(15) << "First offset:" << firstOffset << std::endl;
+      std::cerr << std::setprecision(15) << "Last offset:" << lastOffset << std::endl;
+      std::cerr << std::setprecision(15) << "Next offset " << nextOffset << std::endl;
+
       // We're looking for a zero, so check that either firstLine and middleLine have
       // opposite signs, or middleLine and lastLine have opposite signs.
       if ((firstOffset > 0) == (nextOffset < 0)) {
@@ -580,6 +593,11 @@ csm::ImageCoord UsgsAstroLsSensorModel::groundToImage(
          break;
       }
    }
+
+   std::cerr << std::setprecision(15) << "Final First time:" << firstTime << std::endl;
+   std::cerr << std::setprecision(15) << "Final Last time:" << lastTime << std::endl;
+   std::cerr << std::setprecision(15) << "Final First offset:" << firstOffset << std::endl;
+   std::cerr << std::setprecision(15) << "Final Last offset:" << lastOffset << std::endl;
 
    // Check that the desired precision was met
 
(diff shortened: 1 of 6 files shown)


src/UsgsAstroLsSensorModel.cpp
```diff
@@ -561,57 +528,89 @@ csm::ImageCoord UsgsAstroLsSensorModel::groundToImage(
...
+   double lastOffset = computeViewingPixel(lastTime, ground_pt, adj, pixelPrec/2).line - 0.5;
+
+   // Check if both offsets have the same sign.
+   // This means there is not guaranteed to be a zero.
+   if ((firstOffset > 0) != (lastOffset < 0)) {
+        throw csm::Warning(
+           csm::Warning::IMAGE_COORD_OUT_OF_BOUNDS,
+           "The image coordinate is out of bounds of the image size.",
+           "UsgsAstroLsSensorModel::groundToImage");
+   }
```
jessemapel: This may need to go away? I think that switching to the secant method means we don't need opposite signs.

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -2766,7 +2773,7 @@ std::string UsgsAstroLsSensorModel::constructStateFromIsd(const std::string imag
 
 
    state["m_parameterVals"] = std::vector<double>(NUM_PARAMETERS, 0.0);
-   state["m_parameterVals"][15] = state["m_focal"].get<double>();
+   // state["m_parameterVals"][15] = state["m_focal"].get<double>();
```
jessemapel: We are storing focal length in two places. We should either only store it in one place or only use on in calculations, see line 1733.

tests/CMakeLists.txt
```diff
@@ -3,7 +3,8 @@ cmake_minimum_required(VERSION 3.10)
 # Link runCSMCameraModelTests with what we want to test and the GTest and pthread library
 add_executable(runCSMCameraModelTests
                PluginTests.cpp
-               FrameCameraTests.cpp)
+               FrameCameraTests.cpp
+               LineScanCameraTests.cpp)
```
jessemapel: We should switch to what ISIS does and just glob all of the cpp files in the tests directory.

tests/data/constVelocityLineScan.json
```diff
@@ -3,15 +3,15 @@
...
-    [0.5, 992.0, 0.1]
+    [0.5, -0.8, 0.1]
   ],
   "detector_sample_summing": 1,
   "detector_line_summing": 1,
-  "t0_ephemeris": -8.0,
+  "t0_ephemeris": -0.8,
   "dt_ephemeris": 0.2,
-  "t0_quaternion": -8.0,
+  "t0_quaternion": -0.8,
```
jessemapel: All of these are just errors in the original meta-data.

tests/data/constVelocityLineScan.json
```diff
@@ -23,7 +23,7 @@
   "image_samples": 16,
   "detector_center" : {
     "line" : 0.5,
-    "sample" : 7.5
+    "sample" : 8.0
```
jessemapel: Our CCD is 1 line by 16 samples, so the middle should be 16/2=8, 1/2=0.5

jessemapel:
This changes the ISD spec meaning. detector_center is now expected to be CSM, 0 based pixels. The top left corner of the CCD is (0,0) and pixels are [n,n+1) X [m, m+1).

src/UsgsAstroLsSensorModel.cpp
jlaura: :+1: no need for f(x_{1})f(x_{2}) < 0 for secant.

src/UsgsAstroLsSensorModel.cpp
jlaura: I think that we want it in the parameterVals? If we want to do an adjustment and let the focal length vary, we need to be able to make modifications to the adjustments parameter. I think the only way to vary focal is if it is in the parameterVals vector.

tests/CMakeLists.txt
jlaura: https://cmake.org/cmake/help/v3.13/command/file.html?#filesystem
> 
> Note
> 
> We do not recommend using GLOB to collect a list of source files from your source tree. If no CMakeLists.txt file changes when a source is added or removed then the generated build system cannot know when to ask CMake to regenerate. The CONFIGURE_DEPENDS flag may not work reliably on all generators, or if a new generator is added in the future that cannot support it, projects using it will be stuck. Even if CONFIGURE_DEPENDS works reliably, there is still a cost to perform the check on every rebuild.

tests/LineScanCameraTests.cpp
```diff
@@ -0,0 +1,74 @@
...
+   // EXPECT_NO_THROW(
+   //       sensorModel->replaceModelState(modelState)
+   // );
+
+   // When this is different, the output is very hard to parse
+   // TODO implement JSON diff for gtest
+
+   // EXPECT_EQ(sensorModel->getModelState(), modelState);
+}
+
```
jlaura: what are these?

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -561,57 +528,89 @@ csm::ImageCoord UsgsAstroLsSensorModel::groundToImage(
...
+      auto referenceTimeIt = std::upper_bound(m_intTimeStartTimes.begin(),
+                                              m_intTimeStartTimes.end(),
+                                              nextTime);
+      if (referenceTimeIt != m_intTimeStartTimes.begin()) {
+         --referenceTimeIt;
+      }
+      size_t referenceIndex = std::distance(m_intTimeStartTimes.begin(), referenceTimeIt);
+      double computedLine = (nextTime - m_intTimeStartTimes[referenceIndex]) / m_intTimes[referenceIndex]
+                          + m_intTimeLines[referenceIndex] - 0.5; // subtract 0.5 for ISIS -> CSM pixel conversion
+      double closestLine = floor(computedLine + 0.5);
```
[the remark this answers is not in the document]
jessemapel: We should also pull the conversion in a different PR. At the same time we should change the starting sample/line to 0 based.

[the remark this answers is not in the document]
jessemapel: Because that will need to be coordinated with pfefferneuse

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -2504,9 +2510,10 @@ csm::ImageCoord UsgsAstroLsSensorModel::computeViewingPixel(
 
    // Convert to image sample line
    double line = detectorLine + m_detectorLineOrigin - m_detectorLineOffset
-               - m_offsetLines + 0.5;
-   double sample = (detectorSample + m_detectorSampleOrigin - m_startingSample)
-                 / m_detectorSampleSumming - m_offsetSamples + 0.5;
+               - m_offsetLines;
+   double sample = (detectorSample + m_detectorSampleOrigin - m_startingSample + 1.0)
```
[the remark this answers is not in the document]
jessemapel: Yes see [1694](https://github.com/USGS-Astrogeology/usgscsm/blob/a78ad02ec7b90395a3c1b061e8f517dd139870a0/src/UsgsAstroLsSensorModel.cpp#L1694)

src/UsgsAstroLsSensorModel.cpp
jessemapel: There's 2 things we should do then.

1. Remove explicit focal length stuff from our metadata stuff.
2. Remove m_focal from all calculations and use adjusted parameter 15.

I will do 2 on this PR.

tests/LineScanCameraTests.cpp
jessemapel: I copied tests over from the framer. The state test should work. There's something wrong with how we are writing out and setting state in the line scanner.

The other tests just need their values re-calculated for the line scan test case.

tests/CMakeLists.txt
jessemapel: Is there a better way to manage the list of test sources? This won't be an issue here, but it will be for ISIS.

jlaura:
@jessemapel @kberryUSGS Is this good to merge?  If so, this should auto-rebuild the .dll for @thareUSGS via appveyor.

kberryUSGS:
:+1: This looks good to me.

jessemapel:
This should be ready

Issue #165, opened 2019-02-11
Modularize compute viewing pixel


jlaura:
What are the different pieces that can be tested independently? What are all of the pieces doing? Inline docs.

PR #169, opened 2019-02-12
Initial steps to modularize UsgsAstroLsSensorModel::losToEcf

### LINKED ISSUE #166: Modularize los2Ecf

kberryUSGS:
... along with some of `computeViewingPixel` as a bonus. This PR does not add tests for the newly added functions. Looking for feedback about the way this is modularized first.

Also, notably this doesn't (yet) pull these functions out into a separate file or make any attempt for portability / usability between the linescanner and framer sensor models, though this would be good to do in the future. This should fix #166 and includes steps toward #165.

When this does get merged, please squash and merge. The commit history includes converting back-and-forth between unix-style and dos-style carriage returns and this really doesn't need to be recorded.


jessemapel:
RE moving to a utility file. @acpaquette has already started doing that with distortion. I think we can just do all of that after this gets merged.

include/usgscsm/UsgsAstroLsSensorModel.h
```diff
@@ -916,7 +916,38 @@ class UsgsAstroLsSensorModel : public csm::RasterGM, virtual public csm::Settabl
...
       csm::WarningList* warnings = NULL) const;
 
-   // This method computes the imaging locus.
+   // methods pulled out of los2ecf
+
+   void computeDistortedFocalPlaneCoordinates(
+       const double& lineUSGS, 
+       const double& sampleUSGS, 
+       double &distortedLine, 
+       double& distortedSample) const;
```
jessemapel: We should get rid of all the USGS/isis variable names. I'd much rather see lineCCD, sampleCCD than lineUSGS, sampleUSGS.

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -1666,6 +1652,138 @@ double UsgsAstroLsSensorModel::getValue(
...
+
+    attCorr[0] = cos_b * cos_c;
+    attCorr[1] = -cos_a * sin_c + sin_a * sin_b * cos_c;
+    attCorr[2] = sin_a * sin_c + cos_a * sin_b * cos_c;
+    attCorr[3] = cos_b * sin_c;
+    attCorr[4] = cos_a * cos_c + sin_a * sin_b * sin_c;
+    attCorr[5] = -sin_a * cos_c + cos_a * sin_b * sin_c;
+    attCorr[6] = -sin_b;
+    attCorr[7] = sin_a * cos_b;
+    attCorr[8] = cos_a * cos_b;
```
jessemapel: This is a rotation matrix calculated from euler angles. It would be nice to have a function similar to the one for quaternions that computes a rotation matrix from euler angles.

include/usgscsm/UsgsAstroLsSensorModel.h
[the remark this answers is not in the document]
kberryUSGS: :+1:

include/usgscsm/UsgsAstroLsSensorModel.h
```diff
@@ -916,7 +916,38 @@ class UsgsAstroLsSensorModel : public csm::RasterGM, virtual public csm::Settabl
...
+
+   void computeUndistortedFocalPlaneCoordinates(
+       const double& isisNatFocalPlaneX, 
+       const double& isisNatFocalPlaneY, 
+       double& isisFocalPlaneX, 
+       double& isisFocalPlaneY) const;
```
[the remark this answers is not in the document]
kberryUSGS: :+1:

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -1666,6 +1652,138 @@ double UsgsAstroLsSensorModel::getValue(
...
+      + losIsis[1] * losIsis[1]
+      + losIsis[2] * losIsis[2]);
+   losIsis[0] /= isisMag;
+   losIsis[1] /= isisMag;
+   losIsis[2] /= isisMag;
+};
+
+
+// Given a time and a flag to indicate whether the a->b or b->a rotation should be calculated
+// uses the quaternions in the m_quaternions member to calclate a rotation matrix.
```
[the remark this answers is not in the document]
kberryUSGS: At some point I think we were trying to (or required to?) avoid adding dependencies? If we can, adding a linear algebra package would be great.

src/UsgsAstroLsSensorModel.cpp
kberryUSGS: Sure, this sounds like a good idea, and will do.

src/UsgsAstroLsSensorModel.cpp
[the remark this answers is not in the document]
jessemapel: I'm curious how the team feels about adding a linear algebra package like armadillo @Kelvinrr @acpaquette @jlaura @thareUSGS

include/usgscsm/UsgsAstroLsSensorModel.h
```diff
@@ -919,22 +919,26 @@ class UsgsAstroLsSensorModel : public csm::RasterGM, virtual public csm::Settabl
...
+       double& undistortedFocalPlaneY) const; 
 
    void calculateRotationMatrixFromQuaternions(
        const double& time, 
        const bool& invert, 
        double cameraToBody[9]) const;
 
+   void calculateRotationMatrixFromEuler(
+       double euler[], 
+       double rotationMatrix[]) const;
```
jessemapel: This needs to be a reference to return properly?

src/UsgsAstroLsSensorModel.cpp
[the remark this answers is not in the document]
kberryUSGS: I didn't respond earlier to your other point: I did change this to just calculate the rotation matrix, and updated the caller that used invert=true to apply the transpose.

include/usgscsm/UsgsAstroLsSensorModel.h
SgStapleton: This is how UsgsAstroLsSensorModel::determineSensorCovarianceInImageSpace made use of it's array (sensor_cov) and successfully uses it in the same way (updating it within the method call).

kberryUSGS: @jessemapel Doesn't this implicitly pass a pointer to the first element of the array? It's not exactly the same thing as pass by reference, but doesn't it function similarly in practice?

jessemapel: @kberryUSGS That's right, I forgot about how arrays are passed around.

PR #170, opened 2019-02-12
Finish Modularizing computeViewingPixel

SgStapleton:


src/UsgsAstroLsSensorModel.cpp
```diff
@@ -1808,44 +1861,44 @@ void UsgsAstroLsSensorModel::losToEcf(
    double sampleUSGSFull = sampleCSMFull;
    double fractionalLine = line - floor(line);
 
-   // Compute distorted image coordinates in mm (sample, line on image (pixels) -> focal plane 
-   double isisNatFocalPlaneX, isisNatFocalPlaneY; 
+   // Compute distorted image coordinates in mm (sample, line on image (pixels) -> focal plane
+   double isisNatFocalPlaneX, isisNatFocalPlaneY;
```
SgStapleton: These variables should probably be updated...?

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -1808,44 +1861,44 @@ void UsgsAstroLsSensorModel::losToEcf(
...
 
-   // Compute distorted image coordinates in mm (sample, line on image (pixels) -> focal plane 
-   double isisNatFocalPlaneX, isisNatFocalPlaneY; 
+   // Compute distorted image coordinates in mm (sample, line on image (pixels) -> focal plane
+   double isisNatFocalPlaneX, isisNatFocalPlaneY;
    computeDistortedFocalPlaneCoordinates(fractionalLine, sampleUSGSFull, isisNatFocalPlaneX, isisNatFocalPlaneY);
 
    // Remove lens distortion
-   double isisFocalPlaneX, isisFocalPlaneY; 
+   double isisFocalPlaneX, isisFocalPlaneY;
```
SgStapleton: These variables should probably be updated...?

src/UsgsAstroLsSensorModel.cpp
kberryUSGS: Agreed. Good catch, I missed these.

src/UsgsAstroLsSensorModel.cpp
kberryUSGS: Agreed. Good catch, I missed these.

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -1774,11 +1774,64 @@ void UsgsAstroLsSensorModel::calculateAttitudeCorrection(const double& time, con
...
-  
+
   calculateRotationMatrixFromEuler(euler, attCorr);
 }
 
 
+// This method works by iteratively adding distortion until the new distorted
+// point, r, undistorts to within a tolerance of the original point, rp.
+void UsgsAstroLsSensorModel::reconstructSensorDistortion(
+    double focalX,
```
kberryUSGS: Don't `focalX` and `focalY` need to be passed by reference so that they will be updated by this function? (Also, if the existing tests didn't catch this, I'm guessing they don't test the distortion?)

SgStapleton: Yes they do! That is what I meant to do! Good catch!

kberryUSGS: :+1:

PR #171, opened 2019-02-12
Example for LS testing

jessemapel:


src/UsgsAstroLsSensorModel.cpp
```diff
@@ -2138,19 +2138,6 @@ void UsgsAstroLsSensorModel::losEllipsoidIntersect(
...
-      scale += slope * (height - h);
-      x = xc + scale * xl;
-      y = yc + scale * yl;
-      z = zc + scale * zl;
-      hprev = h;
-      computeElevation(x, y, z, h, aPrec, desired_precision);
-      slope = (sprev - scale) / (hprev - h);
-      ktr++;
-   }
-
```
jessemapel: I'm not sure why this code block is here, but it seems to be an odd iterative approach that is already handled based on the math we do earlier.

jlaura: Is this a hold over form the original author? If height = 0 this intersection should always be computable in a single step? Vector intersection on the ellipsoid shouldn't need an iterative solution, but what if the height provided is above or below the ellipsoid? Is this handling occlusion where the ellipsoid can occlude itself because the provided height is 'under' the surface?

jessemapel: You should be able to just add height to the ellipsoid radii before you do the intersection. That is also solvable analytically.

jessemapel: I don't think we're doing this right now. I may add that in under this PR.

jessemapel: It turns out height is already being added to the radii, so this is ready to go now.

jessemapel:
@kberryUSGS @jlaura This is ready to go now.

kberryUSGS:
@jessemapel :+1:

PR #174, opened 2019-02-14
Pulled losToEcf shared functions out into Utilities class

kberryUSGS:
* Pulled losToEcf shared functions out into Utilities class. 
* Left the functions that Adam is working with in his PR as-is
* No tests yet in this PR
* added a getQuaternion member function to return the quaternion at a specified time (lagrange interpolation happens inside this function.)


src/UsgsAstroFrameSensorModel.cpp
```diff
@@ -125,7 +125,7 @@ csm::ImageCoord UsgsAstroFrameSensorModel::groundToImage(
 
   // Camera rotation matrix
   double m[3][3];
-  calcRotationMatrix(m,adjustments);
+  calcRotationMatrix(m, adjustments);
```
SgStapleton: Thank you for always fixing these silly little things you find!

include/usgscsm/UsgsAstroLsSensorModel.h
```diff
@@ -916,43 +915,25 @@ class UsgsAstroLsSensorModel : public csm::RasterGM, virtual public csm::Settabl
...
-       const double& line,
-       const double& sample,
-       double& distortedLine,
-       double& distortedSample) const;
+   void reconstructSensorDistortion(
+     double& focalX,
+     double& focalY,
+     const double& desiredPrecision) const;
 
    void computeUndistortedFocalPlaneCoordinates(
```
SgStapleton: Why was this one not moved to the Utilities class? Was it just not used elsewhere?

src/UsgsAstroFrameSensorModel.cpp
kberryUSGS: :+1: :)

include/usgscsm/UsgsAstroLsSensorModel.h
kberryUSGS: `reconstructSensorDistortion` is removed and replaced with one of the distortion functions that @acpaquette pulled out in PR #172. I just left it as-is in this PR so that tests would pass.

kberryUSGS:
This could get pulled in as-is as an incremental step if people are comfortable with it.

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -1862,16 +1789,16 @@ void UsgsAstroLsSensorModel::losToEcf(
...
    // Compute distorted image coordinates in mm (sample, line on image (pixels) -> focal plane
-   double natFocalPlaneX, natFocalPlaneY;
-   computeDistortedFocalPlaneCoordinates(fractionalLine, sampleUSGSFull, natFocalPlaneX, natFocalPlaneY);
+   std::tuple<double, double> natFocalPlane; //double natFocalPlaneX, natFocalPlaneY;
+   natFocalPlane = computeDistortedFocalPlaneCoordinates(fractionalLine, sampleUSGSFull, m_detectorSampleOrigin, m_detectorLineOrigin, m_detectorSampleSumming, m_startingSample, m_detectorLineOffset, m_iTransS, m_iTransL);
 
    // Remove lens distortion
    double focalPlaneX, focalPlaneY;
-   computeUndistortedFocalPlaneCoordinates(natFocalPlaneX, natFocalPlaneY, focalPlaneX, focalPlaneY);
+   computeUndistortedFocalPlaneCoordinates(std::get<0>(natFocalPlane), std::get<1>(natFocalPlane), focalPlaneX, focalPlaneY);
```
acpaquette: Is this input, output parameter scheme something we want to keep around?

jessemapel: I'd like to remove it in favor of just returning an object.

kberryUSGS: @jessemapel @acpaquette Sounds good to me. I think this particular function is one that is removed or replaced in @acpaquette's PR #172? Does the same general feedback apply to everything else in here?

jessemapel: I think so, should we do that now, or wait until #172 is done?

kberryUSGS:
Additional questions for standup:

* The Ls Sensor model uses m[3][3] for matricies.
* The Framer Sensor model uses m[9] for matricies.
I'd like to pick one and use it for both. 

There's only a few functions in here and they mostly accept and return arrays. If they should return something else (or accept something else) could we talk about what we'd prefer they return?

kberryUSGS:
@SgStapleton Feel free to pull this in so it can get merged with the tests you started writing. I'm going to work on switching the matrix format and quaternion format discussed in the standup, but that can get merged in later.

PR #175, opened 2019-02-18
Makes GTest fully optional

jlaura:
This PR is trying to make GTest fully optional to fix build issues on the conda-forge feedstock of CSM where GTest is not being installed and the GTest repo is not being cloned.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,9 +1,9 @@
 cmake_minimum_required(VERSION 3.10)
 project(usgscsm VERSION 0.0.1 DESCRIPTION "usgscsm library")
-
+  
+include(GoogleTest)
 include(cmake/gtest.cmake)
 include(GNUInstallDirs)
-include(GoogleTest)
 
 set(CMAKE_CXX_STANDARD 11)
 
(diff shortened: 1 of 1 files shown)


jlaura:
and here is a link to the CI build error this is seeking to fix: https://travis-ci.org/conda-forge/staged-recipes/builds/494648309#L1341

jessemapel:
@jlaura Would the changes required in the build script be made in the conda-forge repo?

jlaura:
Yes. For use locally, simply executing a standard `cmake .. && cmake --build .` then `cd` into the python directory and `python setup.py develop` would be sufficient. Still a little convoluted, but a way to get a clean dev. install working. For an actual release build, we manage the flags on the conda-forge side. I am working on that right now, but it looks like this will end up being:

```
#!/bin/sh
mkdir build && cd build
cmake -D CMAKE_BUILD_TYPE=Release\
      -D CMAKE_INSTALL_PREFIX=$PREFIX\
      -D CMAKE_OSX_DEPLOYMENT_TARGET=10.11\
      $SRC_DIR
cd python
$PYTHON -m pip install . --no-deps -vv
``` 

Does that address your question?

jessemapel:
@jlaura Yes except there is no python in the library. Are you thinking of ALE?

jlaura:
Apologies, thinking of pycsm. That is the lib I am working on now. I'll update: https://github.com/conda-forge/staged-recipes/pull/7559 once we make a 1.1 release at the end of this week. Then we should be good to go with usgscsm via conda-forge.

PR #178, opened 2019-02-18
removes unused setup.py

jlaura:
The purpose of the PR is twofold.

1. Remove the unused setup.py since this library is pure C++.
2. Test that appveyor is back working as expected.


jlaura:
It looks like appveyor just auto fired, so that is working as desired now. The issue there was that the repo was renamed from CSM-CameraModel to usgscsm and this caused appveyor to lose sync.

thareUSGS:
Good find. 🌮  So what if I propose a name change back to CSM-CameraModel? I jest.

Issue #176, opened 2019-02-18
Share ISD parsing code between LS and Framer


jessemapel:
This is largely prompted by #163. The framer has significantly more robust ISD parsing, but we know the linescan instantiates properly in GXP.

Both use the exact same ISD format, so can we share that code between them?

thareUSGS:
While I am hesitant to change this before we fix GXP support, I think we should do it. Especially if we can manage the update in the current sprint.

PR #179, opened 2019-02-18
Refactored ISD parsing into utilities and added warnings

jessemapel:
This adds a lot of copy and paste code in order to support warnings in ISD parsing. I can't think of a cleaner pattern that still provides useful error messages for ISD parsing.

If this is good, a follow up PR to refactor ISD parsing in the LS needs to submitted.

Works towards #176


src/UsgsAstroFrameSensorModel.cpp
```diff
@@ -953,27 +919,37 @@ std::string UsgsAstroFrameSensorModel::constructStateFromIsd(const std::string&
...
+    catch (...) {
+      parsingWarnings->push_back(
+        csm::Warning(
+          csm::Warning::DATA_NOT_AVAILABLE,
+          "Could not compute detector pixel to focal plane coordinate transformation.",
+          "UsgsAstroFrameSensorModel::constructStateFromIsd()"));
+    }
 
-      std::cerr << "Focal To Pixel Transformation Parsed!" << std::endl;
+    std::cerr << "Focal To Pixel Transformation Parsed!" << std::endl;
```
jlaura: This comment is really for all of the `std::cerr` entries:

What are people's thoughts on spdlog? Header only, available on conda-forge, looks to implement many of the nice to have from python's logging module. 

https://github.com/gabime/spdlog
https://anaconda.org/search?q=spdlog

tests/ISDParsingTests.cpp
```diff
@@ -0,0 +1,315 @@
...
+#include <gtest/gtest.h>
+
+#include "Fixtures.h"
+
+TEST(MetricConversion, DistanceConversion) {
+  EXPECT_EQ(1,    metric_conversion(1000, "m", "km"));
+  EXPECT_EQ(1000, metric_conversion(1000, "m", "m"));
+  EXPECT_EQ(1000, metric_conversion(1, "km", "m"));
+  EXPECT_EQ(1,    metric_conversion(1, "km", "km"));
+}
```
jlaura: what about 0 or -1000 or -1000.01? Just wondering on possible edge cases?

tests/ISDParsingTests.cpp
```diff
@@ -0,0 +1,315 @@
...
+  };
+  EXPECT_EQ(2, getDetectorStartingSample(isd));
+}
+
+TEST(ISDParsing, MinHeight) {
+  json isd = {
+    {"reference_height", {
+      {"minheight", -1},
+      {"unit", "km"}}
+    }
```
jlaura: this tests one of the edge cases I was asking about above, but in, arguably, the wrong place.

src/UsgsAstroFrameSensorModel.cpp
```diff
@@ -791,156 +793,120 @@ void UsgsAstroFrameSensorModel::replaceModelState(const std::string& stringState
...
 
-        state["m_ccdCenter"][0] = line;
-        state["m_ccdCenter"][1] = sample;
+    std::cerr << "Sensor Orientation Parsed!" << std::endl;
 
-        std::cerr << "Detector Center Parsed!" << std::endl;
-      }
+    // get optical_distortion
+    state["m_odtX"] = getTransverseDistortionX(isd, parsingWarnings);
+    state["m_odtY"] = getTransverseDistortionY(isd, parsingWarnings);
```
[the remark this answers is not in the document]
jessemapel: Not completely yet. The LS still just does radial and the framer still just does transverse.

src/UsgsAstroFrameSensorModel.cpp
```diff
@@ -791,156 +793,120 @@ void UsgsAstroFrameSensorModel::replaceModelState(const std::string& stringState
...
-
-      state["m_startingDetectorSample"] = isd.at("starting_detector_sample");
-      state["m_startingDetectorLine"] = isd.at("starting_detector_line");
-
-      std::cerr << "Detector Starting Pixel Parsed!" << std::endl;
+    state["m_modelName"] = getSensorModelName(isd, parsingWarnings);
+    state["m_imageIdentifier"] = getImageId(isd, parsingWarnings);
+    state["m_sensorName"] = getSensorName(isd, parsingWarnings);
+    state["m_platformName"] = getPlatformName(isd, parsingWarnings);
+    std::cerr << "Model Name Parsed!" << std::endl;
```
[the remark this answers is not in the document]
jessemapel: What do you mean error checking here? The error checking was moved out of here so that it could be shared between the framer and ls.

src/UsgsAstroFrameSensorModel.cpp
jessemapel: These are hold overs from trying to get GXP working. I'd be up for just removing them.

jlaura: :+1: Let's just pull these then.

src/UsgsAstroFrameSensorModel.cpp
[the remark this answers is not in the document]
jlaura: This logging is GXP error checking. If we are yanking the other items, we should yank these to. If we are keeping them in, we should kick them into a logging framework assuming the cost to do that is low.

src/UsgsAstroFrameSensorModel.cpp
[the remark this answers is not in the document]
jlaura: :+1: Resolving as that is for another PR.

PR #180, opened 2019-02-19
Removed ground to image 10m error

jessemapel:
This error shouldn't be in here. The desired precision should be checking for this type of thing.


kberryUSGS:
@jessemapel Before closing this, if this check (more or less) needs to be added somewhere else as a "precision check" could you put up a ticket to add this check where it belongs?

PR #183, opened 2019-02-19
Fixed state being written out and read in differently

jessemapel:
I chose to make the value just a flat array because that better lines up with how we are handling arrays in the state.


jessemapel:
Partial fix for #120

jessemapel:
@thareUSGS I think this is causing an issue when trying to instantiate the LS in GXP. Looking at the logs, it says

```
2019/02/19 11:28:30 File SenSocCsm30SensorModel.cpp 	Line 2305 : ERROR : ERROR: Unable to find a CSM plugin given Sensor state
```

PR #184, opened 2019-02-20
Distortion Integration

### LINKED ISSUE #167: Implement radial and transverse distortion models in framer and Ls
This should wait until after #165 and #166. We also need to be able to parse the ISD/state to select the correct distortion.

acpaquette:
For real this time. Uses a switch statement to handle all logic in regards to applying distortion models. Closes #167


jlaura:
:+1: Nice job seeing this through.

include/usgscsm/Distortion.h
```diff
@@ -3,23 +3,36 @@
 
 #include <vector>
 #include <math.h>
+#include <cmath>
```
jlaura: why math.h and cmath if math.h is included?

jupyter_notebooks/Untitled.ipynb
```diff
@@ -0,0 +1,83 @@
+{
```
jlaura: ??? I am guessing this goes with a rebase?

jupyter_notebooks/UsgsAstroFrameSensorModelTesting.ipynb
```diff
@@ -670,7 +670,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.6.5"
+   "version": "3.7.1"
```
jlaura: Also gone I am guessing?

src/Distortion.cpp
```diff
@@ -162,93 +83,144 @@ std::tuple<double, double> distortionFunction(double ux, double uy,
...
+                      double desiredPrecision) {
+  ux = dx;
+  uy = dy;
+
+  switch (distortionType) {
+    // Compute undistorted focal plane coordinate given a distorted
+    // coordinate set and the distortion coefficients
+    case RADIAL: {
+      double rr = dx * dx + dy * dy;
+      double tolerance = 1.0E-6;
```
jlaura: Just a for me question. Are these safe values to hard code in here for a tolerance?

src/Distortion.cpp
```diff
@@ -162,93 +83,144 @@ std::tuple<double, double> distortionFunction(double ux, double uy,
...
+      }
+    }
+    break;
+    // Computes undistorted focal plane (x,y) coordinates given a distorted focal plane (x,y)
+    // coordinate. The undistorted coordinates are solved for using the Newton-Raphson
+    // method for root-finding if the distortionFunction method is invoked.
+    case TRANSVERSE: {
+      // Solve the distortion equation using the Newton-Raphson method.
+      // Set the error tolerance to about one millionth of a NAC pixel.
+      const double tol = 1.4E-5;
```
jlaura: Same tolerance question as above.

src/Distortion.cpp
```diff
@@ -162,93 +83,144 @@ std::tuple<double, double> distortionFunction(double ux, double uy,
...
+
+      // The maximum number of iterations of the Newton-Raphson method.
+      const int maxTries = 60;
```
jlaura: Same question here about the constant.

src/UsgsAstroFrameSensorModel.cpp
```diff
@@ -64,8 +64,7 @@ void UsgsAstroFrameSensorModel::reset() {
     m_ccdCenter = std::vector<double>(2, 0.0);
     m_spacecraftVelocity = std::vector<double>(3, 0.0);
     m_sunPosition = std::vector<double>(3, 0.0);
-    m_odtX = std::vector<double>(10, 0.0);
-    m_odtY = std::vector<double>(10, 0.0);
+    m_opticalDistCoeffs.clear();
```
jlaura: :+1: Good to know!

src/Distortion.cpp
jessemapel: This could be replaced with the desired precision.

src/Distortion.cpp
```diff
@@ -162,93 +83,144 @@ std::tuple<double, double> distortionFunction(double ux, double uy,
...
+    case RADIAL: {
+      double rr = dx * dx + dy * dy;
+      double tolerance = 1.0E-6;
+
+      if (rr > tolerance)
+      {
+        double dr = opticalDistCoeffs[0] + (rr * (opticalDistCoeffs[1] + rr * opticalDistCoeffs[2]));
+        ux = dx * (1.0 - dr);
+        uy = dy * (1.0 - dr);
+      }
```
jessemapel: This is the math for adding radial distortion, not removing it.

src/Distortion.cpp
jessemapel: Hard coding a max iterations should be fine. 60 is kind of large number of max iterations. Newton-Raphson should have quadratic convergence. A small number of iterations will get us incredibly close. Something like 20 is probably better.

src/Distortion.cpp
```diff
@@ -162,93 +83,144 @@ std::tuple<double, double> distortionFunction(double ux, double uy,
...
+          drOverR = opticalDistCoeffs[0]
+                 + (r2_prev * (opticalDistCoeffs[1] + (r2_prev * opticalDistCoeffs[2])));
+
+          // Compute new estimate of r
+          r = rp + (drOverR * r_prev);
+          iteration++;
+        }
+        while (fabs(r * (1 - drOverR) - rp) > desiredPrecision);
+        ux = dx / (1.0 - drOverR);
+        uy = dy / (1.0 - drOverR);
```
jessemapel: See my comment about the enumeration. I don't see why both adding and removing radial distortion are in this method.

src/Utilities.cpp
```diff
@@ -547,54 +547,43 @@ double getSemiMinorRadius(json isd, csm::WarningList *list) {
...
+    return coefficients;
   }
   catch (...) {
     if (list) {
       list->push_back(
         csm::Warning(
           csm::Warning::DATA_NOT_AVAILABLE,
           "Could not parse the radial distortion model coefficients.",
-          "Utilities::getRadialDistortion()"));
+          "Utilities::getDistortion()"));
```
jessemapel: This scheme will return a warning if there are radial but not transverse distortion coefficients. That's a problem, because the current logic says an ISD cannot be used if any warnings are returned. Is it possible to return a warning only if neither could be found?

src/Utilities.cpp
```diff
@@ -547,54 +547,43 @@ double getSemiMinorRadius(json isd, csm::WarningList *list) {
...
+    return coefficients;
   }
   catch (...) {
     if (list) {
       list->push_back(
         csm::Warning(
           csm::Warning::DATA_NOT_AVAILABLE,
-          "Could not parse the transverse distortion model Y coefficients.",
-          "Utilities::getTransverseDistortionY()"));
+          "Could not parse a transverse distortion model coefficients.",
```
jessemapel: "a set of transverse distortion model coefficients." The coefficients are plural, so singular a is not grammatically correct here.

src/Distortion.cpp
```diff
@@ -162,93 +83,144 @@ std::tuple<double, double> distortionFunction(double ux, double uy,
...
-  double rr = inFocalPlaneX * inFocalPlaneX + inFocalPlaneY * inFocalPlaneY;
-  std::tuple<double, double> undistortedPoint(inFocalPlaneX, inFocalPlaneY);
-
-  if (rr > tolerance)
-  {
-    double dr = opticalDistCoef[0] + (rr * (opticalDistCoef[1] + rr * opticalDistCoef[2]));
-    undistortedPoint = std::make_tuple(inFocalPlaneX * (1.0 - dr), inFocalPlaneY * (1.0 - dr));
+  for (int i = 0; i < 10; xPointer++, yPointer++, i++) {
+    dx = dx + f[i] * opticalDistCoeffs[xPointer];
+    dy = dy + f[i] * opticalDistCoeffs[yPointer];
```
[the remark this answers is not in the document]
acpaquette: Renamed to computeTransverseDistortion

src/UsgsAstroFrameSensorModel.cpp
```diff
@@ -64,8 +61,8 @@ void UsgsAstroFrameSensorModel::reset() {
     m_ccdCenter = std::vector<double>(2, 0.0);
     m_spacecraftVelocity = std::vector<double>(3, 0.0);
     m_sunPosition = std::vector<double>(3, 0.0);
-    m_odtX = std::vector<double>(10, 0.0);
-    m_odtY = std::vector<double>(10, 0.0);
+    m_distortionType = DistortionType::TRANSVERSE;
+    m_opticalDistCoeffs.clear();
```
jessemapel: This should default to 20 0s. Right now, the transverse distortion method would seg-fault because it expects 20 coefficients.

src/UsgsAstroFrameSensorModel.cpp
```diff
@@ -793,120 +796,150 @@ void UsgsAstroFrameSensorModel::replaceModelState(const std::string& stringState
...
+        json typemap = {
+          {"m", 0},
+          {"km", 3}
+        };
+
+        // everything to lowercase
+        std::transform(from.begin(), from.end(), from.begin(), ::tolower);
+        std::transform(to.begin(), to.end(), to.begin(), ::tolower);
+        return val*pow(10, typemap[from].get<int>() - typemap[to].get<int>());
+    };
```
jessemapel: This code is now in utilities and shouldn't be here too.

src/UsgsAstroFrameSensorModel.cpp
```diff
@@ -793,120 +796,150 @@ void UsgsAstroFrameSensorModel::replaceModelState(const std::string& stringState
...
+      state["m_ephemerisTime"] = isd.at("center_ephemeris_time");
+      state["m_nLines"] = isd.at("image_lines");
+      state["m_nSamples"] = isd.at("image_samples");
+
+      state["m_iTransL"] = isd.at("focal2pixel_lines");
+
+      state["m_iTransS"] = isd.at("focal2pixel_samples");
+
+      // We don't pass the pixel to focal plane transformation so invert the
+      // focal plane to pixel transformation
```
jessemapel: All of this code was un-done with the rebase. This whole chunk except for the distortion specific stuff needs to be reverted.

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -328,70 +331,69 @@ void UsgsAstroLsSensorModel::reset()
...
-  m_ephemRates.clear();                      // 43
-  m_quaternions.clear();                     // 44
+  m_startingEphemerisTime = 0.0;
+  m_centerEphemerisTime = 0.0;
+  m_detectorSampleSumming = 1.0;
+  m_startingSample = 1.0;
+  m_ikCode = -85600;
+  m_focal = 350.0;
+  m_zDirection = 1.0;
+  m_distortionType = (DistortionType)0;
```
jessemapel: This should default to something from the enumeration, not 0 cast to the enumeration.

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -328,70 +331,69 @@ void UsgsAstroLsSensorModel::reset()
...
+  m_opticalDistCoeffs.clear();
```
jessemapel: Same as in framer. Default to no distortion for whatever model is defaulted to.

src/Utilities.cpp
```diff
@@ -547,57 +547,80 @@ double getSemiMinorRadius(json isd, csm::WarningList *list) {
...
-  catch (...) {
-    if (list) {
-      list->push_back(
-        csm::Warning(
-          csm::Warning::DATA_NOT_AVAILABLE,
-          "Could not parse the transverse distortion model X coefficients.",
-          "Utilities::getTransverseDistortionX()"));
-    }
+  else if (distortion.compare("radial") == 0) {
+    return DistortionType::RADIAL;
```
jessemapel: You could use [find](https://nlohmann.github.io/json/classnlohmann_1_1basic__json_a89eb3928f57903677051c80534be9cb1.html#a89eb3928f57903677051c80534be9cb1) here instead of iterating and checking each object

src/Utilities.cpp
```diff
@@ -547,57 +547,80 @@ double getSemiMinorRadius(json isd, csm::WarningList *list) {
...
+      catch (...) {
+        if (list) {
+          list->push_back(
+            csm::Warning(
+              csm::Warning::DATA_NOT_AVAILABLE,
+              "Could not parse the radial distortion model coefficients.",
+              "Utilities::getDistortion()"));
+        }
+      }
+      return coefficients;
```
jessemapel: This should return a warning if it doesn't find any distortion coefficients for radial or transverse

src/Utilities.cpp
```diff
@@ -547,57 +547,80 @@ double getSemiMinorRadius(json isd, csm::WarningList *list) {
...
     }
+    break;
```
jessemapel: This should return a warning if it doesn't find any distortion coefficients for radial or transverse.

This method can also end without returning. It either needs a warning and return after the switch statement, or a warning and return in a default case of the switch statement.

acpaquette:
I have absolutely no idea why this is segfaulting on the invalid camera for framer

src/UsgsAstroFrameSensorModel.cpp
```diff
@@ -849,7 +850,6 @@ std::string UsgsAstroFrameSensorModel::constructStateFromIsd(const std::string&
     // sun position is not strictly necessary, but is required for getIlluminationDirection.
     state["m_sunPosition"] = getSunPositions(isd);
 
-
     // get sensor_orientation quaternion
     std::vector<double> quaternion = getSensorOrientations(isd, parsingWarnings);
     if (!quaternion.empty() && quaternion.size() != 4) {
```
acpaquette: @jessemapel The way this is being parsed now makes the InvalidCamera test for the framer fail via seg fault. Removing the !quaternion.empty check fixes it. Does it matter if the quaternion is empty?

PR #181, opened 2019-02-19
Gtests for Utilities.cpp

SgStapleton:
Tests for Utilities.cpp as well as refactoring of method to return by parameter.

Also, removed unneeded focalBias and halfSwath parameters as per @jessemapel guidance.

PR #186, opened 2019-02-20
Updated LS ISD parsing code to match Framer

### LINKED ISSUE #176: Share ISD parsing code between LS and Framer

Kelvinrr:
Closes #176


src/UsgsAstroLsSensorModel.cpp
```diff
@@ -1698,21 +1663,21 @@ void UsgsAstroLsSensorModel::losToEcf(
...
    // USGS image convention: UL pixel center == (1.0, 1.0)
-   double sampleCSMFull = sample + m_offsetSamples;
+   double sampleCSMFull = sample;
    double sampleUSGSFull = sampleCSMFull;
    double fractionalLine = line - floor(line);
 
    // Compute distorted image coordinates in mm (sample, line on image (pixels) -> focal plane
    std::tuple<double, double> natFocalPlane;
-   natFocalPlane = computeDistortedFocalPlaneCoordinates(fractionalLine, sampleUSGSFull, m_detectorSampleOrigin, m_detectorLineOrigin, m_detectorSampleSumming, m_startingSample, m_detectorLineOffset, m_iTransS, m_iTransL);
+   natFocalPlane = computeDistortedFocalPlaneCoordinates(fractionalLine, sampleUSGSFull, m_detectorSampleOrigin, m_detectorLineOrigin, m_detectorSampleSumming, m_startingSample, 0, m_iTransS, m_iTransL);
```
kberryUSGS: If we're eliminating `m_detectorLineOffset`, can we also just take it out of this function signature and definition?

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -2498,147 +2454,108 @@ double UsgsAstroLsSensorModel::determinant3x3(double mat[9]) const
...
+  // Zero computed state values
+  state["m_referencePointXyz"] = std::vector<double>(3, 0.0);
 
-     for (auto& velocity : jayson.at("velocities")) {
-       state["m_ephemRates"].push_back(metric_conversion(velocity[0].get<double>(), unit));
-       state["m_ephemRates"].push_back(metric_conversion(velocity[1].get<double>(), unit));
-       state["m_ephemRates"].push_back(metric_conversion(velocity[2].get<double>(), unit));
-     }
-   }
+  // leave these be for now.
```
kberryUSGS: :+1: On "leave these be for now" but what is happening with these state keywords now / in the future?

kberryUSGS: As in, is the intent to change or remove them from the state, or do they just need to be temporarily initialized?

Kelvinrr: I imagine we want to support these in the ISD and copy them into the state like the others. Not 100% sure why they are being hardcoded or to what extent do these values actually matter.

src/UsgsAstroLsSensorModel.cpp
Kelvinrr: Probably a better question for @jessemapel

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -140,49 +132,55 @@ void UsgsAstroLsSensorModel::replaceModelState(const std::string &stateString )
    auto j = json::parse(stateString);
    int num_params    = NUM_PARAMETERS;
    int num_paramsSq = num_params * num_params;
+   std::cerr << "A" << std::endl;
```
jessemapel: Debug stuff?

jessemapel: Debug things?

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -140,49 +132,55 @@ void UsgsAstroLsSensorModel::replaceModelState(const std::string &stateString )
...
-   m_totalSamples = j["m_totalSamples"];
-   m_offsetLines = j["m_offsetLines"];
-   m_offsetSamples = j["m_offsetSamples"];
+   m_sensorName = j["m_sensorName"];
+   m_nLines = j["m_nLines"];
+   m_nSamples = j["m_nSamples"];
    m_platformFlag = j["m_platformFlag"];
-   m_aberrFlag = j["m_aberrFlag"];
-   m_atmRefFlag = j["m_atmRefFlag"];
+   std::cerr << "A2" << std::endl;
```
jessemapel: More debug

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -193,11 +191,15 @@ void UsgsAstroLsSensorModel::replaceModelState(const std::string &stateString )
...
+   std::cerr << "D" << std::endl;
+
+   m_positions = j["m_positions"].get<std::vector<double>>();
+   m_velocities = j["m_velocities"].get<std::vector<double>>();
    m_quaternions = j["m_quaternions"].get<std::vector<double>>();
-   m_parameterVals = j["m_parameterVals"].get<std::vector<double>>();
+   m_currentParameterValue = j["m_currentParameterValue"].get<std::vector<double>>();
    m_covariance = j["m_covariance"].get<std::vector<double>>();
+
+   std::cerr << "Unpacked The State JSON" << std::endl;
```
jessemapel: All the debug cerrs should be removed. This can happen here or elsewhere. There are similar debugs in the framer.

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -1104,22 +1082,18 @@ void UsgsAstroLsSensorModel::setParameterCovariance(
 //***************************************************************************
 std::string UsgsAstroLsSensorModel::getTrajectoryIdentifier() const
 {
-   return m_trajectoryIdentifier;
+   return "UKNOWN";
```
jessemapel: UNKNOWN

src/UsgsAstroLsSensorModel.cpp
jessemapel: Image flip flag - can be removed. It's some weird old code that was in the model.
GSD - I don't know what this is
flying height - This is a computed value, so it shouldn't be in the ISD
half time - This is a computed value, so it shouldn't be in the ISD

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -2502,147 +2458,108 @@ double UsgsAstroLsSensorModel::determinant3x3(double mat[9]) const
...
-
-     unit = unit.get<std::string>();
-     state["m_semiMajorAxis"] = metric_conversion(semiminor.get<double>(), unit);
-     state["m_semiMinorAxis"] = metric_conversion(semimajor.get<double>(), unit);
-   }
+  // These are exlusive to LineScanners, leave them here for now.
+  state["m_dtEphem"] = isd.at("dt_ephemeris");
+  state["m_t0Ephem"] = isd.at("t0_ephemeris");
+  state["m_dtQuat"] =  isd.at("dt_quaternion");
+  state["m_t0Quat"] =  isd.at("t0_quaternion");
```
jessemapel: We should add warning methods for these values.

src/UsgsAstroLsSensorModel.cpp
jessemapel: @kberryUSGS That sounds good

Kelvinrr:
The tests are failing partly because a previous merged PR made tests with the StartingSample bug, probably worth trying to knock this out first before merging others.

jessemapel:
@Kelvinrr Are you going to fix those under this PR?

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -420,10 +405,10 @@ void UsgsAstroLsSensorModel::updateState()
    // This routine will set some parameters not found in the ISD.
 
    // Reference point (image center)
-   double lineCtr = m_totalLines / 2.0;
-   double sampCtr = m_totalSamples / 2.0;
+   double lineCtr = m_nLines / 2.0;
+   double sampCtr = m_nSamples / 2.0;
    csm::ImageCoord ip(lineCtr, sampCtr);
-   double refHeight = m_refElevation;
+   double refHeight = 30;
```
jessemapel: This should be changed to 0. It was 30 in reset, but was getting overwritten with 0 when the state was added. This is causing the attitude correction test to fail.

Kelvinrr:
Yes? Look at it at the moment. If Summer or someone else wants to fix them (the ones added by #181) then I wouldn't say no.

jessemapel:
@Kelvinrr If these changes don't fix it, I'm okay with merging and then putting in a new PR to get this fixed, I know you didn't want to work today.

Kelvinrr:
Your suggestion above fixed one of them, the other seems to be fixing by undoing the bug fix. In which case, if no one minds merging this and moving forward, go for it.

Kelvinrr:
I believe I've addressed all the comments.

PR #190, opened 2019-02-22
Merge dev into master for 1.1 release

jessemapel:
Once this is merged we can finally pull the release 🎊


jessemapel:
appveryor is failing on the conda-build step, which we don't care about on here because we're going to be pushing up via conda-forge. If we still want to push here, we can update the meta.yml after the release is pulled off, right now it's still trying to build 1.0

Issue #193, opened 2019-02-22
1.1.0 Release Has GTest Dependency?


jlaura:
Working on the conda-forge build and getting test errors. It looks like gtest.h was added to the framer header.

https://github.com/USGS-Astrogeology/usgscsm/blob/dev/include/usgscsm/UsgsAstroFrameSensorModel.h#L8

This looks like it was done in error?

kberryUSGS:
I can take a look at this.

jlaura:
Here is the git blame line. I bet this is an error and got in with PR for distortion.

`6a1c1e16 include/usgscsm/UsgsAstroFrameSensorModel.h      (Tyler Wilson     2018-08-24 09:45:16 -0700   8) #include <gtest/gtest.h>`

kberryUSGS:
@jlaura It doesn't build with the gtest include taken out. 

All of the friend tests in here cause compilation errors like: 

```
In file included from /home/kberry/csm/CSM-CameraModel/src/UsgsAstroPlugin.cpp:3:
/home/kberry/csm/CSM-CameraModel/include/usgscsm/UsgsAstroFrameSensorModel.h:317:17: error: ‘FramerParameterizedTest’ has not been declared
     FRIEND_TEST(FramerParameterizedTest,JacobianTest);
```

If the gtest include is taken out.

jlaura:
weird - I'll add a gtest dependency to the build.

jlaura:
I am going to reopen this. I am not sure how the conda build was successful before, given that the gtest dependency has been in here, but I can not do a conda build with how we are including test currently. 

I think that we need to add:
```
# Find GTest
find_package(GTest REQUIRED)`
```

To the CMakeLists.txt and then modify the includes as:
```
target_include_directories(usgscsm
                           PUBLIC
                           ${USGSCSM_INCLUDE_DIRS}
                           ${CSM_INCLUDE_DIR}
                           ${GTEST_INCLUDE_DIRS}
)
```

I think though that this will mess with the way that gtest is being included as a submodule. It looks like the submodule is the reason that we have been able to include GTest without actually including it. Using the above method, we would use gtest from conda-forge.

With how we are including gtest right now, I am not sure how conda build has been working at all. Thoughts?

kberryUSGS:
@jlaura I am not sure if this will fix it because I didn't try to do a conda build, but I just put up PR #194 which will build and run tests without error without the gtest submodule present.

[...]

PR #194, opened 2019-02-23
Attempt to fix gtest build problem in #193

kberryUSGS:


kberryUSGS:
Forgot to add gtest to Travis conda environment

kberryUSGS:
If we just decided to remove the FRIEND tests instead by making the tested methods public, this PR can just get closed.

kberryUSGS:
Especially as I'm not sure how to fix the build failure: Linux is building, but then has one test failure, and OSX looks like it had some kind of linking failure.

PR #195, opened 2019-02-25
Removes gtest dependency

jlaura:


jlaura:
This removes the protected friends. As far as I can tell, the protected friend tests did not exist?

PR #197, opened 2019-02-25
Further isolates gtest

jlaura:
So, I did not isolate GTest enough. I just tested this on my Mac and:

```
TEST END: /Users/jlaura/miniconda3/conda-bld/osx-64/usgscsm-1.1.1-0.tar.bz2
Renaming work directory,  /<me>/miniconda3/conda-bld/usgscsm_1551115360642/work  to  /<me>/miniconda3/conda-bld/usgscsm_1551115360642/work_moved_usgscsm-1.1.1-0_osx-64_main_build_loop
# Automatic uploading is disabled
# If you want to upload package(s) to anaconda.org later, type:

anaconda upload /Users/jlaura/miniconda3/conda-bld/osx-64/usgscsm-1.1.1-0.tar.bz2

# To have conda build upload to anaconda.org automatically, use
# $ conda config --set anaconda_upload yes

anaconda_upload is not set.  Not uploading wheels: []
```

PR #198, opened 2019-02-25
only add gtest if tests are being built

Kelvinrr:
Without this change, I wasn't able to build usgscsm without gtest.


Kelvinrr:
lol apparently @jlaura and I made the same change @ #197, I'll close this.

PR #201, opened 2019-03-13
Fixed imageToRemoteImagingLocus in LS and a bug with quaternions

### LINKED ISSUE #200: UsgsAstroLsSensorModel::imageToRemoteImagingLocus appears to be incorrect
Looking at the code for this method it does the following:

```
height_delta = pixel ground scale 
ground_point_1 = imageToGround(image_point, 0.0)
ground_point_2 = imageToGround(image_point, height_delta)

point = ground_point_1
direction = normalized(ground_point_2 - ground_point_1)
```

The direction vector is supposed to be the look vector. That's just a subset of the imageToGround call...

jessemapel:
Fixes #200

### DIFF
--- a/src/UsgsAstroLsSensorModel.cpp
+++ b/src/UsgsAstroLsSensorModel.cpp
@@ -882,37 +882,18 @@ csm::EcefLocus UsgsAstroLsSensorModel::imageToRemoteImagingLocus(
    double* achieved_precision,
    csm::WarningList* warnings) const
 {
-   // Object ray unit direction at exposure station
-   // Exposure station arbitrary for orthos, so set elevation to zero
-
-   // Set esposure station elevation to zero
-   double height = 0.0;
-   double GND_DELTA = m_gsd;
-   // Point on object ray at zero elevation
-   csm::EcefCoord gp1 = imageToGround(
-      image_pt, height, desired_precision, achieved_precision, warnings);
-
-   // Unit vector along object ray
-   csm::EcefCoord gp2 = imageToGround(
-      image_pt, height - GND_DELTA, desired_precision, achieved_precision, warnings);
-
-   double dx2, dy2, dz2;
-   dx2 = gp2.x - gp1.x;
-   dy2 = gp2.y - gp1.y;
-   dz2 = gp2.z - gp1.z;
-   double mag2 = sqrt(dx2 * dx2 + dy2 * dy2 + dz2 * dz2);
-   dx2 /= mag2;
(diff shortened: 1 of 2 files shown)


jlaura:
For whatever reason, I was inspired to go back to the docs:

> There is a one-to-one mapping going from ground space to image space. However, going from image to ground space the mapping is one to many. Each pixel in image space corresponds to a line of sight that traverses a set of ground points. That set of ground points is termed the imaging locus. The imaging locus is useful in a variety of photogrammetric operations. For example, an image-to-ground calculation can be made by finding the intersection of the imaging locus with a ground surface.

> In general, the imaging locus is a complicated path due to the effects of physical phenomena. Therefore, the sensor models return an approximation to the imaging locus termed the proximate imaging locus. The proximate imaging locus consists of a point on the locus together with the direction of the locus at that point. In effect, the proximate imaging locus at a given ground point is the line tangent to the imaging locus at that point. Any arbitrary imaging locus can then be approximated using multiple proximate imaging loci.

> For optical imagery, the imaging locus follows the line of sight from the camera to the ground. Figure 4 depicts the imaging locus for optical imagery. The effects of refraction on the line of sight are greatly exaggerated in this picture. For SAR imagery, the imaging locus follows the curve of intersection of a constant range sphere with the Doppler cone and also includes the effects of refraction.

> In cases where a ground point is not available to compute the proximate imaging locus, a remote imaging locus can be obtained. The remote imaging locus is the line of sight approximated by a straight line. The complicated effects of physical phenomena are ignored. This gives a good approximation of the actual imaging locus and can be used to obtain ground points at which a proximate imaging locus can be computed.

Given all of that, it looks like the code that has been removed does provide the desired functionality for the proximate imaging locus. The deleted is a copy of 824-853.  So, yeah this looks like either (a) we are misunderstanding the documentation (I don't think so...) or (b) the previous implementation was problematic.

I am going to merge because I need to get the quaternion fix in for autocnet.

jessemapel:
I think the proximate imaging locus code is correct. Look at this diagram

https://www.geogebra.org/graphing/chjuvnjk

thareUSGS:
The function was changed for the 3.0.3 version (the last version BAE sent
us to open source.
https://github.com/USGS-Astrogeology/usgscsm/blob/32f1557013d04871ea6e954ada4c14080bdf741e/src/UsgsAstroLsSensorModel.cpp#L653

I think they made it more compliant as Jay alluded too. Not sure what we
need to do (or ask) to resolve this.

-Trent

PR #202, opened 2019-03-18
Fixed reading junk data, also cleaned up groundToImage

### LINKED ISSUE #199: ConstVelocityLineScanSensorModel.Inversion Intermitten Test Failure
Here is the failure. This is intermittent.

```
43: Test command: /Users/jlaura/github/usgscsm/build/tests/runCSMCameraModelTests "--gtest_filter=ConstVelocityLineScanSensorModel.Inversion" "--gtest_also_run_disabled_tests"
43: Test timeout computed to be: 1500
43: Note: Google Test filter = ConstVelocityLineScanSensorModel.Inversion
43: [==========] Running 1 test from 1 test case.
43: [----------] Global test environment set-up.
43: [----------] 1 test from ConstVelocityLineScanSensorModel
43: [ RUN      ] ConstVelocityLineScanSensorModel.Inversion
43: /Users/jlaura/github/usgscsm/tests/LineScanCameraTests.cpp:41: Failure
43: Expected equality of these values:
43:   imagePt.line
43:     Which is: 8.5
43:   imageReprojPt.line
43:     Which is: -7.5808080808080813
43: [  FAILED  ] ConstVelocityLineScanSensorModel.Inversion (4 ms)
43: [----------] 1 test from ConstVelocityLineScanSensorModel (4 ms total)
43:
43: [----------] Global test environment tear-down
43: [==========] 1 test from 1 test case ran. (4 ms total)
43: [  PASSED  ] 0 tests.
43: [  FAILED  ] 1 test, listed below:
43: [  FAILED  ] ConstVelocityLineScanSensorModel.Inversion
43:
43:  1 FAILED TEST
43/97 Test #43: ConstVelocityLineScanSensorModel.Inversion ......................................................................***Failed    0.02 sec
```

jessemapel:
It looks like reading off the end of the array was causing intermittent inaccuracies when interpolating the position and velocity.

Fixes #199 I ran the inversion test a 1000 times before and got failures. After the fix, no more failures!

### DIFF
--- a/src/UsgsAstroLsSensorModel.cpp
+++ b/src/UsgsAstroLsSensorModel.cpp
@@ -479,8 +479,10 @@ csm::ImageCoord UsgsAstroLsSensorModel::groundToImage(
    else {
       --approxNextPoint.line;
    }
-   csm::EcefCoord approxIntersect = imageToGround(approxPoint, 0);
-   csm::EcefCoord approxNextIntersect = imageToGround(approxNextPoint, 0);
+   double height, aPrec;
+   computeElevation(ground_pt.x, ground_pt.y, ground_pt.z, height, aPrec, desired_precision);
+   csm::EcefCoord approxIntersect = imageToGround(approxPoint, height);
+   csm::EcefCoord approxNextIntersect = imageToGround(approxNextPoint, height);
    double lineDX = approxNextIntersect.x - approxIntersect.x;
    double lineDY = approxNextIntersect.y - approxIntersect.y;
    double lineDZ = approxNextIntersect.z - approxIntersect.z;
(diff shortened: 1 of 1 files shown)


src/UsgsAstroLsSensorModel.cpp
```diff
@@ -479,8 +479,10 @@ csm::ImageCoord UsgsAstroLsSensorModel::groundToImage(
    else {
       --approxNextPoint.line;
    }
-   csm::EcefCoord approxIntersect = imageToGround(approxPoint, 0);
-   csm::EcefCoord approxNextIntersect = imageToGround(approxNextPoint, 0);
+   double height, aPrec;
+   computeElevation(ground_pt.x, ground_pt.y, ground_pt.z, height, aPrec, desired_precision);
+   csm::EcefCoord approxIntersect = imageToGround(approxPoint, height);
+   csm::EcefCoord approxNextIntersect = imageToGround(approxNextPoint, height);
```
jlaura: This is only tangentially related to this PR, but when the computeLinearApproximation is called above, where/how is the _linear variable being set to not just default to the center of the image?

I am asking because, are we starting close for the secant method to be able to not hit a potential local 0?

jessemapel: replaceModelState attempts to set the linear approximation. We don't actually use this linear approximation as an input into the secant method search. It's possible we should to save an iteration or two.

jessemapel: The main use for the approximate point is to convert the desiredPrecision from ground distance to image distance.

PR #212, opened 2019-04-03
Framer Fix

acpaquette:
Fixes framer and also updates logging to look at the same sink


include/usgscsm/UsgsAstroFrameSensorModel.h
```diff
@@ -14,6 +14,8 @@
 #include "spdlog/sinks/basic_file_sink.h"
 
 #include <json/json.hpp>
+
+#define MESSAGE_LOG(logger, ...) if (logger) { logger->info(__VA_ARGS__); }
```
jessemapel: Move this define into the source file. If it's in the header it is available elsewhere and has much higher chance of colliding with another define from some other library.

include/usgscsm/UsgsAstroFrameSensorModel.h
```diff
@@ -311,6 +312,7 @@ class UsgsAstroFrameSensorModel : public csm::RasterGM {
         const GeometricModel &comparisonModel,
         csm::param::Set pSet = csm::param::VALID,
         const GeometricModelList &otherModels = GeometricModelList()) const;
+    virtual std::shared_ptr<spdlog::logger> getLogger();
```
jessemapel: Why is this virtual? Could also be inlined

PR #222, opened 2019-04-11
Added pixel summing in framer

### LINKED ISSUE #220: Framer does not support summing
The framer needs to support binning/summing. This is important for MDIS and many other instruments.

jessemapel:
Fixes #220 

- Changed the focal plane coordinate computation to just use 2 doubles instead of a tuple so that is better matches the rest of the utility functions.
- Removed the unused principal point from the framer.
- Added a utility function to convert a focal plane coordinate into a pixel. I didn't modify the line scanner to use this because it can't convert all the way to an image line. I'm also not sure how to incorporate line summing in the line scanner because it assumes 1 by X pixel ccds instead of full push broom sensors.
- Added more tests for utilities and the framer

### DIFF
--- a/src/UsgsAstroFrameSensorModel.cpp
+++ b/src/UsgsAstroFrameSensorModel.cpp
@@ -1235,11 +1235,9 @@ void UsgsAstroFrameSensorModel::losEllipsoidIntersect(
 {
    MESSAGE_LOG(this->m_logger, "Calculating losEllipsoidIntersect with height: {},\n\
                                 xc: {}, yc: {}, zc: {}\n\
-                                xl: {}, yl: {}, zl: {}\n\
-                                x: {}, y: {}, z{}", height,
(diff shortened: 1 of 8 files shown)


src/UsgsAstroFrameSensorModel.cpp
```diff
@@ -1282,6 +1272,9 @@ void UsgsAstroFrameSensorModel::losEllipsoidIntersect(
    x = xc + scale * xl;
    y = yc + scale * yl;
    z = zc + scale * zl;
+
+   MESSAGE_LOG(this->m_logger, "Calculated losEllipsoidIntersect at: {},, {}, {}",
```
SgStapleton: Small typo. Extra comma.

SgStapleton:
From what I could follow on the logic, this looked good.

PR #224, opened 2019-04-12
Fixed ground partials in framer and added a test

### LINKED ISSUE #223: Framer computeGroundPartials is incorrect
It looks like it's computing the partials of the focal plane coordinate (and not the pixels).

jessemapel:
Fixes #223

### DIFF
--- a/src/UsgsAstroFrameSensorModel.cpp
+++ b/src/UsgsAstroFrameSensorModel.cpp
@@ -568,6 +568,15 @@ std::vector<double> UsgsAstroFrameSensorModel::computeGroundPartials(const csm::
   double dz = nextPoint.z - z;
   double pixelGroundSize = sqrt((dx*dx + dy*dy + dz*dz) / 2.0);
 
+  // If the ground size is too small, try the opposite direction
+  if (pixelGroundSize < 1e-10) {
+    nextPoint = imageToGround(csm::ImageCoord(ipB.line - 1, ipB.samp - 1));
+    dx = nextPoint.x - x;
+    dy = nextPoint.y - y;
+    dz = nextPoint.z - z;
+    pixelGroundSize = sqrt((dx*dx + dy*dy + dz*dz) / 2.0);
+  }
+
   csm::ImageCoord ipX = groundToImage(csm::EcefCoord(x + pixelGroundSize, y, z));
   csm::ImageCoord ipY = groundToImage(csm::EcefCoord(x, y + pixelGroundSize, z));
   csm::ImageCoord ipZ = groundToImage(csm::EcefCoord(x, y, z + pixelGroundSize));
--- a/src/UsgsAstroFrameSensorModel.cpp
+++ b/src/UsgsAstroFrameSensorModel.cpp
@@ -43,12 +43,12 @@ void UsgsAstroFrameSensorModel::reset() {
     m_focalLength = 0.0;
     m_startingDetectorSample = 0.0;
     m_startingDetectorLine = 0.0;
+    m_detectorSampleSumming = 1.0;
+    m_detectorLineSumming = 1.0;
     m_targetName = "";
     m_ifov = 0;
     m_instrumentID = "";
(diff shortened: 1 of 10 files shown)


src/UsgsAstroFrameSensorModel.cpp
```diff
@@ -840,7 +817,13 @@ void UsgsAstroFrameSensorModel::replaceModelState(const std::string& stringState
...
         m_collectionIdentifier = state.at("m_collectionIdentifier").get<std::string>();
         // Set reference point to the center of the image
-        m_referencePointXyz = imageToGround(csm::ImageCoord(m_nLines / 2.0, m_nSamples / 2.0));
+        m_referencePointXyz = imageToGround(csm::ImageCoord(m_nLines/2.0, m_nSamples/2.0));
+        // Compute the pixel ground size at the reference point
+        csm::EcefCoord nextRefPoint = imageToGround(csm::ImageCoord(m_nLines/2.0 + 1, m_nSamples/2.0 + 1));
+        double dx = nextRefPoint.x - m_referencePointXyz.x;
+        double dy = nextRefPoint.y - m_referencePointXyz.y;
+        double dz = nextRefPoint.z - m_referencePointXyz.z;
+        m_pixelGroundSize = sqrt((dx*dx + dy*dy + dz*dz) / 2.0);
```
jessemapel: It is worth noting that this is not going to be accurate for highly oblique images and most flyby/disk resolved images. We can compute it more precisely for a specific groundpoint/imagepoint but I am mirroring the implementation from the LS here.

jessemapel: @twilson271828 fixed this in ISIS and has a good write up on an [issue](https://github.com/USGS-Astrogeology/ISIS3/issues/2205) there

scsides: It is worth noting. We have been asked by a few groups and scientists here to implement a more accurate resolution. That is exactly why the new one was done.

src/UsgsAstroFrameSensorModel.cpp
```diff
@@ -77,6 +77,7 @@ void UsgsAstroFrameSensorModel::reset() {
     m_referencePointXyz.x = 0;
     m_referencePointXyz.y = 0;
     m_referencePointXyz.z = 0;
+    m_pixelGroundSize = 1.0;
```
scsides: Why are we setting this to 1.0 in what amounts to the constructor?  Would it be better to set it to 0.0?

jessemapel: This value is used in division, so I defaulted it to 1 to avoid division by 0 problems.

src/UsgsAstroFrameSensorModel.cpp
jessemapel: It sounds like it may be better to calculate it at the specific pixel location instead of once for the whole image then.

I'll work on a generic pixel resolution method

src/UsgsAstroFrameSensorModel.cpp
Kelvinrr: It's being used as a denominator later, so I imagine to avoid divide by zero errors. Probably should have a more explicit way of dealing with it.

jessemapel:
@scsides @Kelvinrr Ready for re-review now. I decided not to use the formula that ISIS uses, and instead just directly measure the ground pixel size.

src/UsgsAstroFrameSensorModel.cpp
scsides: Usually if a variable is used when it was not set correctly, it is better to fail ungracefully then to blindly move forward with a bad calculation. The problem presents its self much earlier with a catastrophic failure (instantly) then with a wrong answer at the end (possible never). @Kelvinrr an I talked about this, is it time to consider how best to handle errors within usgs csm?

scsides:
That's one way around the issue

PR #225, opened 2019-04-13
Adding distortion for Kaguya TC

Kelvinrr:
100% jacked from ISIS (https://github.com/USGS-Astrogeology/ISIS3/blob/kaguyatc/isis/src/kaguya/objs/KaguyaTcCamera/KaguyaTcCameraDistortionMap.cpp)


src/Distortion.cpp
```diff
@@ -173,6 +158,47 @@ void removeDistortion(double dx, double dy, double &ux, double &uy,
...
+      //   INS<INSTID>_DISTORTION_COEF_Y  = ( b0, b1, b2, b3),
+      //
+      // Distance r from the center:
+      //   r = - (n - INS<INSTID>_CENTER) * INS<INSTID>_PIXEL_SIZE.
+      //
+      // Line-of-sight vector v is calculated as
+      //   v[X] = INS<INSTID>BORESIGHT[X]
+      //          +a0 +a1*r +a2*r^2 +a3*r^3 ,
+      //   v[Y] = INS<INSTID>BORESIGHT[Y]
+      //          +r +a0 +a1*r +a2*r^2 +a3*r^3 ,
```
kberryUSGS: I need to change this in the comments here in ISIS, too. After much discussion, we ended up deciding that the second equation should actually be: 

```
//   v[Y] = INS<INSTID>BORESIGHT[Y]
//           b0 +b1*r +b2*r^2 +b3*r^3 
```
so the comments should probably be updated with that.

src/Distortion.cpp
```diff
@@ -233,5 +259,69 @@ void applyDistortion(double ux, double uy, double &dx, double &dy,
...
+      for (int i = 0; i < 50; i++) {
+        xx = xt * xt;
+        yy = yt * yt;
+        rr = xx + yy;
+        r = sqrt(rr);
+        rrr = rr * r;
+
+        // Radial distortion
+        // dr is the radial distortion contribution
+        dr_x = odkx[0] + odkx[1] * r + odkx[2] * rr + odkx[3] * rrr; // why did hayabusa have a -1
```
kberryUSGS: I also forgot to take the comment about "why did hayabusa have a -1" out of the ISIS code.

src/Distortion.cpp
```diff
@@ -233,5 +259,69 @@ void applyDistortion(double ux, double uy, double &dx, double &dy,
...
+        dr_y = odky[0] + odky[1] * r + odky[2] * rr + odky[3] * rrr; // why did hayabusa have a -1
```
kberryUSGS: Same here.

src/Utilities.cpp
```diff
@@ -827,6 +830,30 @@ std::vector<double> getDistortionCoeffs(json isd, csm::WarningList *list) {
...
+      }
+      catch (...) {
+        if (list) {
+          list->push_back(
+            csm::Warning(
+              csm::Warning::DATA_NOT_AVAILABLE,
+              "Could not parse a set of transverse distortion model coefficients.",
+              "Utilities::getDistortion()"));
+        }
+        coefficients = std::vector<double>(20, 0.0);
```
jessemapel: This should default to the no distortion coefficients fro the TC distortion model.

jessemapel: This should default to the no distortion coefficients fro the TC distortion model.

Kelvinrr: Good catch. Should be taken care of.

PR #226, opened 2019-04-17
Logging added to Ls; couple typo corrections in Framer

ladoramkershner:


src/UsgsAstroLsSensorModel.cpp
```diff
@@ -124,6 +127,8 @@ const csm::param::Type
 
 void UsgsAstroLsSensorModel::replaceModelState(const std::string &stateString )
 {
+   MESSAGE_LOG(this->m_logger, "Replacing modle state")
```
kberryUSGS: modle -> model

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -1710,7 +1803,7 @@ void UsgsAstroLsSensorModel::lightAberrationCorr(
    if (1.0 == fabs(cosThetap))
    {
       dxl = 0.0;
-      dyl = 0.0;
+      dyl = 0.0;   MESSAGE_LOG(this->m_logger, "getImageStart")
```
jessemapel: This log should be on a separate line.

ladoramkershner:
@jessemapel and @kberryUSGS  I am headed out for the day, I plan on inserting a little more logging to the front half of the cpp file. I added the exception logging Jesse made to the Framing Camera Model, but I may have broke something in the process. I wanted to give everyone a chance to check out the logging I am doing in case they wanted to give feedback ("more", "less", or "don't forget to log this function").  

Apparently I broke Travis, I am not too practiced (aka never looked at it at all) with Travis, so I will look into this Monday, but may require some assistance. 

Still feel free to give feedback about the level of logging happening!

jessemapel:
The logging looks good to me. That Travis error looks interesting, we may have our mac build not using the write c++.

jessemapel:
CI is failing due to https://github.com/gabime/spdlog/issues/1067, waiting on a response from psdlog folks, we may need to accept a broken mac build in the short term? @all should we just move ahead with this and accept mac CI is broken or wait to merge more changes until spdlog gets libfmt updated?

Kelvinrr:
My vote to just accept macOS CI as broken until the issue is fixed. The only exception is if we can figure out why it was working until now and somehow bring it back to that state (e.g. did Travis update its macOS version? In which case can we explicitly change the version for now? https://docs.travis-ci.com/user/reference/osx/#macos-version).

Kelvinrr:
My vote to just accept macOS CI as broken until the issue is fixed. The only exception is if we can figure out why it was working until now and somehow bring it back to that state (e.g. did Travis update its macOS version? In which case can we explicitly change the version for now? https://docs.travis-ci.com/user/reference/osx/#macos-version).

jessemapel:
This failure does appear to be related to this PR, I re-opened #229 to re-run the CI and it passed there.

PR #227, opened 2019-04-18
Added tests to check that the framer logs

jessemapel:
Once logging is in the line scan, we should do this for the linescan.

I also added logging to the methods in the framer that throw an exception because they are not supported.

There is one method that is not tested, `getCrossCovarianceMatrix`, because I am unsure how we are supposed to call it.

### DIFF
--- a/include/usgscsm/UsgsAstroFrameSensorModel.h
+++ b/include/usgscsm/UsgsAstroFrameSensorModel.h
@@ -313,6 +313,7 @@ class UsgsAstroFrameSensorModel : public csm::RasterGM {
         csm::param::Set pSet = csm::param::VALID,
         const GeometricModelList &otherModels = GeometricModelList()) const;
     virtual std::shared_ptr<spdlog::logger> getLogger();
+    virtual void setLogger(std::shared_ptr<spdlog::logger> logger);
 
     static const std::string _SENSOR_MODEL_NAME;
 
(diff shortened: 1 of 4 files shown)


SgStapleton:
I noticed there were a handful of functions in UsgsAstroFrameSensorModel.cpp with logging information that were not being covered by these tests. Was this done on purpose?

jessemapel:
@SgStapleton Which methods are those?

SgStapleton:
These were the ones I found:
 * getIlluminationDirection
 * computeGroundPartials
 * isValidModelState
 * isValidIsd
 * constructStateFromIsd
 * getCrossCovarianceMatrix
 * calcRotationMatrix (two methods by this name)
 * losEllipsoidIntersect
 * getValue

jessemapel:
@SgStapleton I've added tests for `getIlluminationDirection` and `computeGroundPartials`. `isValidModelState`, `isValidIsd`, `constructStateFromIsd`, `calcRotationMatrix`, `losEllipsoidIntersect`, and `getValue` are all out own custom methods and not part of the RasterGM API, so I didn't test them. I can add them if you think they should still be tested. For `getCrossCovarianceMatrix` I'm not even sure how to properly call that method right now, so I left it out.

SgStapleton:
@jessemapel I honestly don't know that I know enough to have a qualified opinion as far as what needs to be tested. It seems like this PR was specifically for RasterGM, though. So, if those tests do need to be added, maybe that needs to happen in a separate PR?

jessemapel:
@SgStapleton For coverage's sake, I'll add tests for the non-RasterGM methods because we probably want to test them too.

SgStapleton:
@jessemapel Sounds good. I will be keeping an eye on this PR.

SgStapleton:
I did not see this before I approved - sorry. But were we intending to remove all of the FrameStateTests...?

jessemapel:
@SgStapleton I moved them so that they are all together now. They were split in half by the adjustable parameter tests.

SgStapleton:
Oh my goodness! I'm blind!

PR #228, opened 2019-04-18
Made framer inherit from settable ellipsoid

jessemapel:
See discussion on #142


jessemapel:
GXP is successfully showing stereo and proper footprints for MDIS!

PR #231, opened 2019-04-24
Dawn Distortion

### LINKED ISSUE #230: Add DawnFC Distortion Model
Add the dawn distortion model to the available distortion models. Currently dawnfc defaults to radial with only one coefficient which results in some interesting footprints.

acpaquette:
Adds the dawn distortion model and fixes the line scanners warning system to properly propagate warnings back to the user. Closes #230

### DIFF
--- a/src/Distortion.cpp
+++ b/src/Distortion.cpp
@@ -105,7 +105,6 @@ void removeDistortion(double dx, double dy, double &ux, double &uy,
     // Computes undistorted focal plane (x,y) coordinates given a distorted focal plane (x,y)
     // coordinate. The undistorted coordinates are solved for using the Newton-Raphson
     // method for root-finding if the distortionFunction method is invoked.
-    case KAGUYATC:
     case TRANSVERSE: {
       // Solve the distortion equation using the Newton-Raphson method.
       // Set the error tolerance to about one millionth of a NAC pixel.
@@ -257,7 +256,6 @@ void applyDistortion(double ux, double uy, double &dx, double &dy,
       }
     }
     break;
-    case KAGUYATC:
     case TRANSVERSE: {
       computeTransverseDistortion(ux, uy, dx, dy, opticalDistCoeffs);
     }
--- a/src/Distortion.cpp
+++ b/src/Distortion.cpp
@@ -200,6 +200,62 @@ void removeDistortion(double dx, double dy, double &ux, double &uy,
       ux = dx + dr_x;
       uy = dy + dr_y;
     }
+    break;
+    // The dawn distortion model is "reversed" so the remove function computes
+    // the distorted focal plane but stores it into the undistorted variables (ux, uy)
+    case DAWNFC: {
+      double offsetSqrd;
+      int    numAttempts;
+      double delta;
+      bool    done;
+
+      /****************************************************************************
+      * Pre-loop intializations
+      ****************************************************************************/
+
+      numAttempts = 1;
+      delta = 0.00001;
+
+      offsetSqrd = dy * dy + dx * dx;
+      double guess_dx, guess_dy;
+      double guess_ux, guess_uy;
+
+      /****************************************************************************
+      * Loop ...
+      ****************************************************************************/
+      do {
+        guess_ux = dx / (1.0 + opticalDistCoeffs[0] * offsetSqrd);
+        guess_uy = dy / (1.0 + opticalDistCoeffs[0] * offsetSqrd);
+
+        offsetSqrd = guess_uy * guess_uy + guess_ux * guess_ux;
+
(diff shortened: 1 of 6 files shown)


jlaura:
Check this out: https://www.internalpointers.com/post/squash-commits-into-one-git

src/Distortion.cpp
```diff
@@ -324,5 +395,16 @@ void applyDistortion(double ux, double uy, double &dx, double &dy,
         dy = ydistorted;
       }
     }
+    break;
+    // The dawn distortion model is "reversed" so the apply function computes
+    // the distorted focal plane and stores it into the distorted variables (dx, dy)
```
kberryUSGS: Like above, I understand the situation, but found this comment confusing. The DAWNFC case under applyDistortion still calculates the distorted coefficients using the undistorted coefficients, it just doesn't need to use an iterative solution to do so.

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -2420,8 +2420,21 @@ std::string UsgsAstroLsSensorModel::constructStateFromIsd(const std::string imag
...
+  }
+
+  delete parsingWarnings;
+  parsingWarnings = nullptr;
 
-   // The state data will still be updated when a sensor model is created since
-   // some state data is notin the ISD and requires a SM to compute them.
-   return state.dump();
+  // The state data will still be updated when a sensor model is created since
+  // some state data is notin the ISD and requires a SM to compute them.
```
kberryUSGS: notin -> not in

src/Distortion.cpp
```diff
@@ -200,6 +215,62 @@ void removeDistortion(double dx, double dy, double &ux, double &uy,
       ux = dx + dr_x;
       uy = dy + dr_y;
     }
+    break;
+    // The dawn distortion model is "reversed" so the remove function computes
```
kberryUSGS: I understand the situation, but found this comment confusing. The `DAWNFC` case under `removeDistortion` uses an iterative solution to calculate the undistorted coordinates, given the distorted coordinates. I wouldn't say that it it computes one thing but stores it in the other's variables? 

The iterative solution is necessary here because Dawn's distortion equation is given as distorted = fn(undistorted), which is reversed from the more normal undistorted = fn(distorted).

src/UsgsAstroFrameSensorModel.cpp
```diff
@@ -593,6 +593,9 @@ std::vector<double> UsgsAstroFrameSensorModel::computeGroundPartials(const csm::
   MESSAGE_LOG(this->m_logger, "Computing ground partials results:\nLine: {}, {}, {}\nSample: {}, {}, {}",
                                partials[0], partials[1], partials[2], partials[3], partials[4], partials[5]);
 
+  MESSAGE_LOG(this->m_logger, "Computing ground partials result line: {}, sample: {}, wrt: {} and x: {}, y: {}, z: {}",
+                               partials[0], partials[1], partials[2], partials[3], partials[4], partials[5]);
+
```
jessemapel: This looks like a duplicate line

src/Distortion.cpp
```diff
@@ -200,6 +200,64 @@ void removeDistortion(double dx, double dy, double &ux, double &uy,
       ux = dx + dr_x;
       uy = dy + dr_y;
     }
+    break;
+    // The dawn distortion model is "reversed" from other distortion models so
+    // the remove function iteratively computes undistorted coordinates based on
+    // the distorted coordinates, rather than iteratively computing distorted coordinates
+    // to undistorted coordinates.
+    case DAWNFC: {
+      double offsetSqrd;
```
jessemapel: normally we call this the radius squared

src/Distortion.cpp
```diff
@@ -200,6 +200,64 @@ void removeDistortion(double dx, double dy, double &ux, double &uy,
...
+        guess_dy = guess_uy * (1.0 + opticalDistCoeffs[0] * offsetSqrd);
+
+        done = true;
+        if(abs(guess_dy - dy) > delta) {
+          done = false;
+        }
+
+        if(abs(guess_dx - dx) > delta) {
+          done = false;
+        }
```
jessemapel: usually you want to check if your under tolerance and then set done to true. This is a little confusing because done will be set to true and then immediately set to false 90% of the time. You can condense this into a single && statement and it will even short circuit for a tiny savings.

kberryUSGS: I think this was copied from ISIS, so it may need to be changed there, too?

jessemapel: This is a very minor thing, I wouldn't worry about it in ISIS. It's unusual, but not incorrect.

PR #232, opened 2019-04-25
Distortion Model Testing

acpaquette:
Adds tests for the dawn and kaguya distortion models.

### DIFF
--- a/tests/DistortionTests.cpp
+++ b/tests/DistortionTests.cpp
@@ -199,34 +199,19 @@ TEST(DawnFc, testZeroCoeffs) {
   EXPECT_NEAR(uy,10.0,1e-8);
 }
 
-TEST(KaguyaTc, testApply) {
-  csm::ImageCoord imagePt(2.0, 2.0);
-
-  double dx, dy;
-  double desiredPrecision = 0.0000001;
-  std::vector<double> distortionCoeffs = {1, 0, 0, 0,
-                                          1, 0, 0, 0};
-
-  applyDistortion(imagePt.samp, imagePt.line, dx, dy, distortionCoeffs,
-                  DistortionType::KAGUYATC, desiredPrecision);
-
-  EXPECT_NEAR(dx, 1, 1e-8);
-  EXPECT_NEAR(dy, 1, 1e-8);
-}
-
-TEST(KaguyaTc, testRemove) {
+TEST(KaguyaTc, testRemoveCoeffs) {
   csm::ImageCoord imagePt(1.0, 1.0);
 
   double ux, uy;
   double desiredPrecision = 0.0000001;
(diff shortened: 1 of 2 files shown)


tests/DistortionTests.cpp
```diff
@@ -177,3 +178,92 @@ TEST(DawnFc, testRemove) {
...
+  EXPECT_NEAR(dy, 1, 1e-8);
+}
+
+TEST(KaguyaTc, testRemove) {
+  csm::ImageCoord imagePt(1.0, 1.0);
+
+  double ux, uy;
+  double desiredPrecision = 0.0000001;
+  std::vector<double> distortionCoeffs = {1, 0, 0, 0,
+                                          1, 0, 0, 0};
```
jessemapel: We're only checking that the first term is being handled correctly, we should test non-zero coefficients for the other terms too.

acpaquette: I'm not sure how necessary it is to actually test the individual coefficients like this. The two tests bellow are applying and removing distortion with actual coefficients and all zero coefficients. Both of those seem to be covering all of our bases

tests/DistortionTests.cpp
```diff
@@ -177,3 +178,79 @@ TEST(DawnFc, testRemove) {
...
+                  DistortionType::DAWNFC, desiredPrecision);
+
+  removeDistortion(dx, dy, ux, uy, coeffs,
+                  DistortionType::DAWNFC, desiredPrecision);
+
+
+  EXPECT_NEAR(dx,10.0,1e-8);
+  EXPECT_NEAR(dy,10.0,1e-8);
+  EXPECT_NEAR(ux,10.0,1e-8);
+  EXPECT_NEAR(uy,10.0,1e-8);
```
Kelvinrr: Any reason you are not using `ASSERT_DOUBLE_EQ`

tests/DistortionTests.cpp
```diff
@@ -160,20 +160,97 @@ TEST(DawnFc, testApply) {
...
+  applyDistortion(imagePt.samp, imagePt.line, dx, dy, coeffs,
+                  DistortionType::KAGUYATC, desiredPrecision);
+
+  removeDistortion(dx, dy, ux, uy, coeffs,
+                  DistortionType::KAGUYATC, desiredPrecision);
+
+  EXPECT_NEAR(dx, 0.999566, 1e-6);
+  EXPECT_NEAR(dy, 1.00137, 1e-5);
+  EXPECT_NEAR(ux, 1.0, 1e-8);
+  EXPECT_NEAR(uy, 1.0, 1e-8);
```
Kelvinrr: Why weren't some able to be changed?

PR #242, opened 2019-08-02
removed line sample range check for framer during image time computation

Kelvinrr:


kberryUSGS:
Why remove the check entirely, rather than fix the incorrect comparison?

Kelvinrr:
the time is the same anywhere on the image, doing a range check is just annoying especially when campt also works off image.

jessemapel:
CSM should work in as broad a context as possible, so we should be able to give a time for any pixel location, even off the image.

PR #243, opened 2019-10-25
updated distortion to closer match ISIS

Kelvinrr:
Increased max iteration count, loosened precision, and acceptance criteria for Radial Distortion. 

Isis uses pixel pitch as the tolerance, CSM doesn't have access to pixel pitch so just punting it for now.

### DIFF
--- a/include/usgscsm/Distortion.h
+++ b/include/usgscsm/Distortion.h
@@ -30,5 +30,5 @@ void removeDistortion(double dx, double dy, double &ux, double &uy,
 void applyDistortion(double ux, double uy, double &dx, double &dy,
                      const std::vector<double> opticalDistCoeffs,
                      DistortionType distortionType,
-                     const double desiredPrecision = 1.0E-2, const double tolerance = 1.0E-6);
+                     const double desiredPrecision = 1.0E-6, const double tolerance = 1.0E-6);
 #endif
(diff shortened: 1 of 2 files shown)


src/Distortion.cpp
```diff
@@ -93,13 +93,13 @@ void removeDistortion(double dx, double dy, double &ux, double &uy,
   switch (distortionType) {
 
     // Compute undistorted focal plane coordinate given a distorted
-    // coordinate set and the distortion coefficients
+    // co
```
jessemapel: This looks a typo

include/usgscsm/Distortion.h
```diff
@@ -30,5 +30,5 @@ void removeDistortion(double dx, double dy, double &ux, double &uy,
 void applyDistortion(double ux, double uy, double &dx, double &dy,
                      const std::vector<double> opticalDistCoeffs,
                      DistortionType distortionType,
-                     const double desiredPrecision = 0, const double tolerance = 1.0E-6);
+                     const double desiredPrecision = 1.0E-6, const double tolerance = 1.0E-6);
```
jessemapel: I would make this one argument, the tolerance. The desired precision should be handled by the model that calls this and converted into a convergence tolerance.

kberryUSGS:
Like Stuart, I also agree with what Jesse said.

Kelvinrr:
For now this is basically a bug fix, we should probably talk more about how to properly solve this problem here: https://github.com/USGS-Astrogeology/usgscsm/issues/244

PR #246, opened 2019-10-30
Updated Kaguya LISM distortion model

jessemapel:
This is the distortion model used by both the TC and MI, which are part of the LISM.

The only change from the old Kaguya TC distortion model is accounting for the boresight as indicated in the IKs and Hiro's script.

This is going to require a coordinated change in ALE to provide the updated distortion. I thought about keeping it backwards compatible and setting the boresight values to 0, but I don't want to introduce a situation where the code silently falls back to a less correct solution to support old ISDs.

### DIFF
--- a/src/Distortion.cpp
+++ b/src/Distortion.cpp
@@ -99,7 +99,7 @@ void removeDistortion(double dx, double dy, double &ux, double &uy,
 
       if (rr > tolerance) {
         double dr = opticalDistCoeffs[0] + (rr * (opticalDistCoeffs[1] + rr * opticalDistCoeffs[2]));
-        
+
         ux = dx * (1.0 - dr);
         uy = dy * (1.0 - dr);
       }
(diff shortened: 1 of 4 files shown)


src/Distortion.cpp
```diff
@@ -265,14 +271,20 @@ void removeDistortion(double dx, double dy, double &ux, double &uy,
     case LROLROCNAC: {
 
       if (opticalDistCoeffs.size() != 1) {
-        throw "Distortion coefficients for LRO LROC NAC must be of size 1, current size: " +  std::to_string(opticalDistCoeffs.size());
+        csm::Error::ErrorType errorType = csm::Error::INDEX_OUT_OF_RANGE;
```
jessemapel: I'm not 100% sure this is the correct error type, but it seemed the most appropriate.

src/Distortion.cpp
```diff
@@ -265,14 +271,20 @@ void removeDistortion(double dx, double dy, double &ux, double &uy,
...
+        std::string function = "removeDistortion";
+        throw csm::Error(errorType, message, function);
       }
 
       double dk1 = opticalDistCoeffs[0];
 
       double den = 1 + dk1 * dy * dy;     // r = dy*dy = distance from the focal plane center
       if (den == 0.0) {
-        throw "Unable to remove distortion for LRO LROC NAC. Focal plane position " + std::to_string(dy);
+        csm::Error::ErrorType errorType = csm::Error::ALGORITHM;
```
jessemapel: Same as above, this seems to be the most appropriate error as it's a division by 0 error.

src/Distortion.cpp
```diff
@@ -185,13 +186,18 @@ void removeDistortion(double dx, double dy, double &ux, double &uy,
...
+        std::string function = "removeDistortion";
+        throw csm::Error(errorType, message, function);
       }
 
-      const double* odkx = opticalDistCoeffs.data();
-      const double* odky = opticalDistCoeffs.data()+4;
+      double boresightX = opticalDistCoeffs[0];
+      std::vector<double> odkx(opticalDistCoeffs.begin()+1, opticalDistCoeffs.begin()+5);
+      double boresightY = opticalDistCoeffs[5];
+      std::vector<double> odky(opticalDistCoeffs.begin()+6, opticalDistCoeffs.begin()+10);
```
jessemapel: Switched to use std vectors and range constructors for better error checking. We've already had problems with reading off the end of C arrays. So, I figure it's safer to move to std vectors here because the input data is already a std vector.

tests/DistortionTests.cpp
```diff
@@ -199,57 +199,58 @@ TEST(DawnFc, testZeroCoeffs) {
...
   // Coeffs obtained from file TC1W2B0_01_05211N095E3380.img
-  std::vector<double> coeffs = {-0.0009649900000000001, 0.00098441, 8.5773e-06, -3.7438e-06,
-                                -0.0013796, 1.3502e-05, 2.7251e-06, -6.193800000000001e-06};
+  std::vector<double> coeffs = {-0.0725, -0.0009649900000000001, 0.00098441, 8.5773e-06, -3.7438e-06,
+                                0.0214, -0.0013796, 1.3502e-05, 2.7251e-06, -6.193800000000001e-06};
+
+  removeDistortion(imagePt.samp, imagePt.line, ux, uy, coeffs,
+                  DistortionType::KAGUYALISM, desiredPrecision);
+  applyDistortion(ux, uy, dx, dy, coeffs,
+                  DistortionType::KAGUYALISM, desiredPrecision);
```
jessemapel: I swapped the order here so that I could hand calculate ux and uy. Before it did the iterative solution first which made hand calculating test data challenging.

src/Distortion.cpp
Kelvinrr: `SENSOR_MODEL_NOT_CONSTRUCTIBLE` might be a reasonable alternative?

Issue #236, opened 2019-05-06
Don't set valid image range based on samples and lines in image     


kberryUSGS:
The valid image range should be larger due to distortion models and / or anything else that requires a valid image -> ground outside the original range of the image.

kberryUSGS:
Need to come up with a method of determining what is valid for an image range.

oleg-alexandrov:
The frame and linescan image-to-ground and ground-to-image transforms already handle lines and samples outside the nominal range. The formulas apply as usual, and the results degrade smoothly, becoming unreasonable and then wild the further one goes from the field of view. There is no sharp cutoff where they stop being valid.

getValidImageRange just reports the image box, and nothing internal enforces it. A consumer that needs points outside it, such as ASP during bundle adjustment, can evaluate there without trouble. This is also why the open question of what to report as the valid range has no clean answer. Validity degrades continuously, so any enlarged number would be arbitrary.

Hard range checks have proven counterproductive in the past, as they force bundle adjustment to stop rather than letting it gently compensate for or points that momentarily project just outside and it prevents filtering them properly too. So likely there is not much to do here.

PR #247, opened 2019-11-01
Changed line scan code to zero based CSM instead of ISIS based 0.5

scsides:


src/UsgsAstroLsSensorModel.cpp
```diff
@@ -750,14 +749,15 @@ csm::ImageCoord UsgsAstroLsSensorModel::groundToImage(
...
    double preSquare = desired_precision * desired_precision;
    if (warnings && (desired_precision > 0.0) && (preSquare < len)) {
-      std::stringstream msg;
-      msg << "Desired precision not achieved. ";
-      msg << len << "  " << preSquare << "\n";
-      warnings->push_back(
-         csm::Warning(csm::Warning::PRECISION_NOT_MET,
-         msg.str().c_str(),
-         "UsgsAstroLsSensorModel::groundToImage()"));
+     MESSAGE_LOG(m_logger, "groundToImage: Desired precision not achieve {}", preSquare)
```
jessemapel: missing the d on achieved

jessemapel: Also, this should be logged even when the warning list is not passed by the caller. I'd move the creation of the working to a nested if block. Also, you can re-use the logging string in the warning.

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -1295,28 +1295,19 @@ std::string UsgsAstroLsSensorModel::getReferenceDateAndTime() const
...
-                                           lineUSGSFull);
+                                           lineFull);
    if (referenceLineIt != m_intTimeLines.begin()) {
       --referenceLineIt;
    }
    size_t referenceIndex = std::distance(m_intTimeLines.begin(), referenceLineIt);
 
    double time = m_intTimeStartTimes[referenceIndex]
-      + m_intTimes[referenceIndex] * (lineUSGSFull - m_intTimeLines[referenceIndex]);
+      + m_intTimes[referenceIndex] * (lineFull - m_intTimeLines[referenceIndex] + 0.5);
```
jessemapel: I'd add a comment that adding 0.5 results in getting the center exposure time for the line.

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -1622,7 +1613,7 @@ UsgsAstroLsSensorModel::getValidImageRange() const
 {
    return std::pair<csm::ImageCoord, csm::ImageCoord>(
       csm::ImageCoord(0.0, 0.0),
-      csm::ImageCoord(m_nLines, m_nSamples));
+      csm::ImageCoord(m_nLines, m_nSamples)); // Technically nl and ns are outside the image in a zero based system.
```
jessemapel: per the CSM API docs, this should be the range over which the sensor model is valid and not just the physical image. The API docs say we should return =- 99,999 since there is no exact limit.

The true limit on this would be based on the distortion model, which is hard to generalize.

tests/LineScanCameraTests.cpp
```diff
@@ -131,15 +131,15 @@ TEST_F(OrbitalLineScanSensorModel, Center) {
 }
 
 TEST_F(OrbitalLineScanSensorModel, Inversion) {
-  for (double line = 0.5; line < 16; line++) {
+  for (double line = 0.5; line < 1.1; line+=0.5) {
```
jessemapel: This needs to be put back to 16

src/UsgsAstroLsSensorModel.cpp
acpaquette: @jessemapel Should this function be changed to reflect that the valid range is more dynamic? I guess I'm not sure exactly what should be done about this function in this PR specifically

jessemapel: For right now, we should return the -99,999 default. We could eventually do some computations to find the domain over which the distortion functions are invertible, but that would be way out of scope on this.

jessemapel:
[Finish Stuart's PR](https://trello.com/c/k6lT6gSD/9-finish-stuarts-pr)

src/UsgsAstroLsSensorModel.cpp
jessemapel: This change is outside the scope of this PR though, see #236

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -748,8 +748,8 @@ csm::ImageCoord UsgsAstroLsSensorModel::groundToImage(
    }
 
    double preSquare = desired_precision * desired_precision;
+   MESSAGE_LOG(m_logger, "groundToImage: Desired precision not achieved {}", preSquare)
    if (warnings && (desired_precision > 0.0) && (preSquare < len)) {
-     MESSAGE_LOG(m_logger, "groundToImage: Desired precision not achieve {}", preSquare)
```
jessemapel: This log warning needs to be moved back inside of this if statement.

PR #254, opened 2020-01-07
Removed warning when log_file missing in LS

### LINKED ISSUE #252: Remove warning when log_file is not in the ISD
The GSET people have asked us to remove the warning when no log_file is passed to the ISD.

They have also recommended using an environment variable for logging. This would be a good longer-term solution because it would allow for better logging in the Plugin. We could generate one log file for the plugin and then dynamically generate log files for each sensor model object as they are created. This is fairly easy with our logging library.

jessemapel:
Fixes #252


jessemapel:
[Remove warnings when optional parameters are not present in the ISD](https://trello.com/c/EJs2OrvU/7-remove-warnings-when-optional-parameters-are-not-present-in-the-isd)

PR #256, opened 2020-01-08
Inaccurate Implementation for getReferenceDateAndTime

### LINKED ISSUE #250: Add getReferenceDateAndTime to LS sensor model
This is part of the CSM API that we don't support right now. It should return the UTC date and time for the reference point.

acpaquette:
Adds an inaccurate implementation to get the date and time of a given reference point on an image. This method is inaccurate due to the differences in time computation between NAIF and default timing libraries. Mainly, NAIF accounting for leap seconds in it's timing the the differences between epochs (NAIF: January 1st, 2000 C++: Janurary 1st: 1970)

Closes #250


jessemapel:
[Implement getReferenceDateAndTime in the LS](https://trello.com/c/nLBCrCtn/6-implement-getreferencedateandtime-in-the-ls)

PR #257, opened 2020-01-08
Added methods for computing linearized detector coordinates

jessemapel:
The plane methods would be needed for CPPS search instead of the generalized detector distance, but I left them in if we every want CPPS.


jessemapel:
@acpaquette This should be good to go for the search implementation

jessemapel:
[Implement CPPS computation](https://trello.com/c/ZTBqz9M2/11-implement-cpps-computation)

acpaquette:
@jessemapel Did this get PR'd into the wrong branch?

jessemapel:
@acpaquette looks like I did. I'll PR it into dev too

PR #259, opened 2020-01-10
Added getIlluminationDirection for line scanners

AustinSanders:
I could use some guidance on how to better meet best practices for testing.

One specific concern is that the each test uses multiple functions that aren't intended to be tested in the specific unit (i.e. the getIlluminationDirectionLagrange test uses imageToGround, lagrangeInterp, etc.).


src/UsgsAstroLsSensorModel.cpp
```diff
@@ -443,6 +447,28 @@ std::string UsgsAstroLsSensorModel::getModelState() const {
...
+      state["m_sunVelocity"] = json();
+      state["m_sunVelocity"][0] = m_sunVelocity[0];
+      state["m_sunVelocity"][1] = m_sunVelocity[1];
+      state["m_sunVelocity"][2] = m_sunVelocity[2];
+      MESSAGE_LOG(m_logger, "m_sunVelocity: {} "
+                            "m_sunVelocity: {} "
+                            "m_sunVelocity: {} ",
+                             m_sunVelocity[0],
+                             m_sunVelocity[1],
+                             m_sunVelocity[2])
```
jessemapel: These should append based on the size of the vectors. Just directly adding the first position and first velocity could cause part of the ISD to be ignored.

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -1632,14 +1662,48 @@ UsgsAstroLsSensorModel::getValidImageRange() const
...
+
+  // If there are multiple positions, use Lagrange interpolation
+  if ((numSunPositions/3) > 1) {
+    double sunPos[3];
+    lagrangeInterp(numSunPositions/3, &m_sunPosition[0], m_t0Ephem, m_dtEphem,
+                   imageTime, 3, 8, sunPos);
+    x = groundPt.x - sunPos[0];
+    y = groundPt.y - sunPos[1];
+    z = groundPt.z - sunPos[2];
+  } else {
```
jessemapel: We do else on a new line:

```
if {
}
else {
}
```

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -1632,14 +1662,48 @@ UsgsAstroLsSensorModel::getValidImageRange() const
...
-      "UsgsAstroLsSensorModel::getIlluminationDirection");
+
+  csm::ImageCoord imagePt = groundToImage(groundPt);
+  double imageTime = getImageTime(imagePt);
+
+  short numSunPositions = m_sunPosition.size();
+  short numSunVelocities = m_sunVelocity.size();
+  double x;
+  double y;
+  double z;
```
jessemapel: You can just make an EcefVector here and set its x, y, z instead of creating these.

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -1632,14 +1662,48 @@ UsgsAstroLsSensorModel::getValidImageRange() const
...
+      y = groundPt.y - (imageTime * m_sunVelocity[1] + m_sunPosition[1]);
+      z = groundPt.z - (imageTime * m_sunVelocity[2] + m_sunPosition[2]);
+    // If there is one position triple with no velocity triple, then the
+    //  illumination direction is the difference of the original vectors.
+    } else {
+      x = groundPt.x - m_sunPosition[0];
+      y = groundPt.y - m_sunPosition[1];
+      z = groundPt.z - m_sunPosition[2];
+    }
+  }
```
jessemapel: This should be

```
if (...) {
}
else if (...) {
}
else {
}

PR #260, opened 2020-01-11
New groundToImage method

acpaquette:
Still rough around the edges but I believe this is working as expected. Primarily splits up the logic in computeViewingPixel to access the detector coords and not have to keep applying distortion to the detector coords.


src/UsgsAstroLsSensorModel.cpp
```diff
@@ -1306,7 +1283,7 @@ double UsgsAstroLsSensorModel::getImageTime(
    const csm::ImageCoord& image_pt) const
 {
    // Remove 0.5 after ISIS dependency in the linescanrate is gone
```
jessemapel: Move this comment down to when the actual time is computed. That 0.5 is what needs to be removed once we change the linescan rate so that it is 0 based instead of 0.5 based.

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -654,112 +652,91 @@ csm::ImageCoord UsgsAstroLsSensorModel::groundToImage(
...
+                         + m_iTransS[1] * distortedFocalX
+                         + m_iTransS[2] * distortedFocalY;
 
-   return calculatedPixel;
+   // Convert to image sample line
+   line += detectorLine + m_detectorLineOrigin;
+   double sample = (detectorSample + m_detectorSampleOrigin - m_startingSample)
+                 / m_detectorSampleSumming;
+   MESSAGE_LOG(m_logger, "computeViewingPixel: image line sample {} {}",
+                                line, sample)
```
jessemapel: For test-ability, I would make these two steps a separate function that we can test in isolation.

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -654,112 +688,122 @@ csm::ImageCoord UsgsAstroLsSensorModel::groundToImage(
...
+   while(abs(detectorLine) > desiredPrecision) {
+     if (++count > 15) {
+       break;
+     }
+     timei = getImageTime(approxPt);
+     finalDetectorView = computeDetectorView(timei, groundPt, adj, desiredPrecision);
+
+     // Convert to detector line and sample
+     detectorLine = m_iTransL[0]
+                  + m_iTransL[1] * finalDetectorView[0];
```
jessemapel: The semi-colon at the end of this line is potentially messing things up

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -704,112 +702,159 @@ csm::ImageCoord UsgsAstroLsSensorModel::groundToImage(
...
+             "UsgsAstroLsSensorModel::groundToImage()"));
+       }
+       break;
+     }
+     timei = getImageTime(approxPt);
+     detectorView = computeDetectorView(timei, groundPt, adj, desiredPrecision);
+
+     // Convert to detector line and sample
+     detectorLine = m_iTransL[0]
+                  + m_iTransL[1] * detectorView[0];
```
jessemapel: Still need to remove this semi-colon.

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -704,112 +702,159 @@ csm::ImageCoord UsgsAstroLsSensorModel::groundToImage(
...
+   //   pi = computeDetectorView(timei, groundPt, adj, desiredPrecision);
+   //   di = distanceToLine(pi[0], pi[1], a, b, c);
+   // }
+   //
+   // if (count >= 30) {
+   //   std::cout << "COULDN'T CONVERGE" << '\n';
+   // }
+
+   std::vector<double> detectorView;
+   double detectorLine = 1;
```
jessemapel: Make this like the number of lines in the image. I could see a time when the desired precision is 1 pixel.

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -704,112 +702,159 @@ csm::ImageCoord UsgsAstroLsSensorModel::groundToImage(
...
+                + m_iTransL[2] * distortedFocalY;
+   detectorSample = m_iTransS[0]
+                  + m_iTransS[1] * distortedFocalX
+                  + m_iTransS[2] * distortedFocalY;
+
+   // Convert to image sample line
+   approxPt.line += detectorLine;
+   approxPt.samp = (detectorSample + m_detectorSampleOrigin - m_startingSample)
+                 / m_detectorSampleSumming;
+   achievedPrecision = &detectorLine;
```
jessemapel: This could be NULL, so this is needs an if not Null check.

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -704,112 +702,159 @@ csm::ImageCoord UsgsAstroLsSensorModel::groundToImage(
...
-   double preSquare = desired_precision * desired_precision;
-   if (warnings && (desired_precision > 0.0) && (preSquare < len)) {
-     MESSAGE_LOG(m_logger, "groundToImage: Desired precision not achieved {}", preSquare)
-     std::stringstream msg;
-     msg << "Desired precision not achieved. ";
-     msg << len << "  " << preSquare << "\n";
-     warnings->push_back(csm::Warning(csm::Warning::PRECISION_NOT_MET,
-                         msg.str().c_str(),
-                         "UsgsAstroLsSensorModel::groundToImage()"));
+   if (warnings && (desiredPrecision > 0.0) && (*achievedPrecision > desiredPrecision))
```
jessemapel: Same here, needs a check that `achievedPrecision` is not NULL. You could use short-circuiting here by just adding `&& (achievedPrecision)` in the middle of these two.

tests/LineScanCameraTests.cpp
```diff
@@ -38,8 +38,8 @@ TEST_F(ConstVelocityLineScanSensorModel, Inversion) {
   csm::EcefCoord groundPt = sensorModel->imageToGround(imagePt, 0.0);
   csm::ImageCoord imageReprojPt = sensorModel->groundToImage(groundPt);
 
-  EXPECT_DOUBLE_EQ(imagePt.line, imageReprojPt.line);
-  EXPECT_DOUBLE_EQ(imagePt.samp, imageReprojPt.samp);
+  EXPECT_NEAR(imagePt.line, imageReprojPt.line, 1e-2);
+  EXPECT_NEAR(imagePt.samp, imageReprojPt.samp, 1e-2);
```
jessemapel: Sorry I was wrong here, default precision is 1e-1:

```
   virtual csm::ImageCoord groundToImage(
      const csm::EcefCoord& groundPt,
      const std::vector<double> &adjustments,
      double desiredPrecision = 0.001,
      double* achievedPrecision = NULL,
      csm::WarningList* warnings = NULL) const;
```

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -704,112 +666,76 @@ csm::ImageCoord UsgsAstroLsSensorModel::groundToImage(
 
    // Convert the ground precision to pixel precision so we can
    // check for convergence without re-intersecting
```
jessemapel: These comments are now obsolete and should be removed.

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -704,112 +666,76 @@ csm::ImageCoord UsgsAstroLsSensorModel::groundToImage(
...
-        firstTime = nextTime;
-        firstOffset = nextOffset;
-      }
-      else {
-        lastTime = nextTime;
-        lastOffset = nextOffset;
-      }
-      MESSAGE_LOG(m_logger, "groundToImage: Loop firstOffset {}, firstTime {}, lastOffset {}, lastTime {}, nextOffset {}, nextTime {}",
-                  firstOffset, firstTime, lastOffset, lastTime, nextOffset, nextTime)
+     // Convert to detector line and sample
```
jessemapel: Convert to detector line ~and sample~

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -704,112 +666,76 @@ csm::ImageCoord UsgsAstroLsSensorModel::groundToImage(
...
-                                              nextTime);
-      if (referenceTimeIt != m_intTimeStartTimes.begin()) {
-         --referenceTimeIt;
-      }
+   csm::ImageCoord approxPt;
+   computeLinearApproximation(groundPt, approxPt);
+
+   std::vector<double> detectorView;
+   double detectorLine = m_nLines;
+   double detectorSample;
```
jessemapel: This should be defined closer to where it is used, after the while loop.

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -704,112 +666,76 @@ csm::ImageCoord UsgsAstroLsSensorModel::groundToImage(
...
+
+   while(abs(detectorLine) > desiredPrecision) {
+     if (++count > 15) {
+       if (warnings) {
+         warnings->push_back(
+           csm::Warning(
+             csm::Warning::PRECISION_NOT_MET,
+             "Unable to achieve desired precision within 15 iterations.",
+             "UsgsAstroLsSensorModel::groundToImage()"));
+       }
```
jessemapel: I think we can actually remove this warning here because it will happen later on.

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -704,112 +666,76 @@ csm::ImageCoord UsgsAstroLsSensorModel::groundToImage(
...
-     std::stringstream msg;
-     msg << "Desired precision not achieved. ";
-     msg << len << "  " << preSquare << "\n";
-     warnings->push_back(csm::Warning(csm::Warning::PRECISION_NOT_MET,
-                         msg.str().c_str(),
-                         "UsgsAstroLsSensorModel::groundToImage()"));
+   MESSAGE_LOG(m_logger, "computeViewingPixel: image line sample {} {}",
+                                approxPt.line, approxPt.samp)
+
+   if (warnings && (desiredPrecision > 0.0) && (achievedPrecision) && (*achievedPrecision > desiredPrecision))
```
jessemapel: If the user doesn't pass in an achievedPrecisions, this should still happen. So, I would actually change it to check that `(abs(detectorLine) > desiredPrecision)`

Sorry about sending you on the wrong path with my last review :(

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -2438,16 +2364,17 @@ void UsgsAstroLsSensorModel::getAdjSensorPosVel(
 //***************************************************************************
 // UsgsAstroLineScannerSensorModel::computeViewingPixel
```
jessemapel: This need to be updated to the new name

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -2438,16 +2364,17 @@ void UsgsAstroLsSensorModel::getAdjSensorPosVel(
 //***************************************************************************
 // UsgsAstroLineScannerSensorModel::computeViewingPixel
 //***************************************************************************
-csm::ImageCoord UsgsAstroLsSensorModel::computeViewingPixel(
+std::vector<double> UsgsAstroLsSensorModel::computeDetectorView(
    const double& time,
    const csm::EcefCoord& groundPoint,
    const std::vector<double>& adj,
    const double& desiredPrecision) const
```
jessemapel: desiredPrecision can be removed as an argument because this doesn't do anything iterative anymore.

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -704,112 +666,76 @@ csm::ImageCoord UsgsAstroLsSensorModel::groundToImage(
...
-   double preSquare = desired_precision * desired_precision;
-   if (warnings && (desired_precision > 0.0) && (preSquare < len)) {
-     MESSAGE_LOG(m_logger, "groundToImage: Desired precision not achieved {}", preSquare)
-     std::stringstream msg;
-     msg << "Desired precision not achieved. ";
-     msg << len << "  " << preSquare << "\n";
-     warnings->push_back(csm::Warning(csm::Warning::PRECISION_NOT_MET,
-                         msg.str().c_str(),
-                         "UsgsAstroLsSensorModel::groundToImage()"));
+   MESSAGE_LOG(m_logger, "computeViewingPixel: image line sample {} {}",
```
jessemapel: This should have the function name updated

tests/LineScanCameraTests.cpp
```diff
@@ -34,12 +34,15 @@ TEST_F(ConstVelocityLineScanSensorModel, Center) {
 }
 
 TEST_F(ConstVelocityLineScanSensorModel, Inversion) {
+  double *achievedPrecision = new double;
```
jessemapel: You can just pass a pointer to a local variable here, no need to deal with heap allocation.

tests/LineScanCameraTests.cpp
```diff
@@ -34,12 +34,15 @@ TEST_F(ConstVelocityLineScanSensorModel, Center) {
...
   csm::ImageCoord imagePt(8.5, 8);
   csm::EcefCoord groundPt = sensorModel->imageToGround(imagePt, 0.0);
-  csm::ImageCoord imageReprojPt = sensorModel->groundToImage(groundPt);
+  csm::ImageCoord imageReprojPt = sensorModel->groundToImage(groundPt, 0.001, achievedPrecision);
 
-  EXPECT_DOUBLE_EQ(imagePt.line, imageReprojPt.line);
-  EXPECT_DOUBLE_EQ(imagePt.samp, imageReprojPt.samp);
+  EXPECT_NEAR(*achievedPrecision, 0.001, 1e-1);
```
jessemapel: This should check that achievedPrecision is less than 0.001

tests/LineScanCameraTests.cpp
```diff
@@ -34,12 +34,15 @@ TEST_F(ConstVelocityLineScanSensorModel, Center) {
...
+  EXPECT_NEAR(imagePt.line, imageReprojPt.line, 1e-1);
+  EXPECT_NEAR(imagePt.samp, imageReprojPt.samp, 1e-1);
```
jessemapel: Check tolerance should be 1e-3

tests/LineScanCameraTests.cpp
```diff
@@ -194,22 +170,25 @@ TEST_F(OrbitalLineScanSensorModel, getSunPositionStationary){
...
-    csm::ImageCoord imageReprojPt = sensorModel->groundToImage(groundPt);
+    csm::ImageCoord imageReprojPt = sensorModel->groundToImage(groundPt, 0.001, achievedPrecision);
 
     // groundToImage has a default precision of 0.001m and each pixel is 100m
     // so we should be within 0.1 pixels
+    EXPECT_NEAR(*achievedPrecision, 0.001, 1e-1);
     EXPECT_NEAR(imagePt.line, imageReprojPt.line, 0.1);
     EXPECT_NEAR(imagePt.samp, imageReprojPt.samp, 0.1);
   }
+  delete achievedPrecision;
```
jessemapel: See comments on ConstVelocityLineScanSensorModel.Inversion, they apply here too.

tests/LineScanCameraTests.cpp
jessemapel: Again, I typed the wrong number before :(

PR #262, opened 2020-01-13
Made ground partial step larger

jessemapel:
The ground partials are computed using finite differences. In some cases, the step size was too small to actually difference across multiple pixels which resulted in bad results. I bumped the step size up to 3 times the reference pixel resolution to help fix this. I also added a LS test.


jessemapel:
I don't know if this works to 100% fix the problem. Don't merge this yet.

PR #264, opened 2020-01-14
Made LS image time continuous

jessemapel:
We were having issues with ground partials being inconsistent across the image, this makes them consistent.

### DIFF
--- a/tests/LineScanCameraTests.cpp
+++ b/tests/LineScanCameraTests.cpp
@@ -105,7 +105,7 @@ TEST_F(ConstVelocityLineScanSensorModel, calculateAttitudeCorrection) {
   double attCorr[9];
   adj.resize(15, 0);
   // Pi/2 with simply compensating for member variable m_flyingHeight in UsgsAstroLsSensorModel
-  adj[7] = (M_PI / 2) * 990.0496255790623081338708;
+  adj[7] = (M_PI / 2) * sensorModel->m_flyingHeight;
   sensorModel->calculateAttitudeCorrection(999.5, adj, attCorr);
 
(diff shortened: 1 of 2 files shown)


tests/LineScanCameraTests.cpp
```diff
@@ -238,7 +238,7 @@ TEST_F(OrbitalLineScanSensorModel, InversionHeight) {
 TEST_F(OrbitalLineScanSensorModel, InversionReallyHigh) {
   for (double line = 0.5; line < 16; line++) {
     csm::ImageCoord imagePt(line, 8);
-    csm::EcefCoord groundPt = sensorModel->imageToGround(imagePt, 49000.0);
```
jessemapel: This was too close to the sensor, it caused the ground to image to diverge because of how close it was.

tests/LineScanCameraTests.cpp
```diff
@@ -105,7 +105,7 @@ TEST_F(ConstVelocityLineScanSensorModel, calculateAttitudeCorrection) {
   double attCorr[9];
   adj.resize(15, 0);
   // Pi/2 with simply compensating for member variable m_flyingHeight in UsgsAstroLsSensorModel
-  adj[7] = (M_PI / 2) * 990.0496255790623081338708;
+  adj[7] = (M_PI / 2) * sensorModel->m_flyingHeight;
```
acpaquette: Why this change? I guess, why was it hard coded in the first place?

jessemapel: This was just hard-coded when the test was written, I think we forgot the sensor model state was public. I just happened to break it, so I made it more flexible.

jessemapel:
This is obsoleted by #260 So, I'm closing it. If we ever have problems with that we can re-investigate this.

PR #266, opened 2020-01-15
Fixed orbital ls

jessemapel:
This fixes the orbital LS test model so that it images 100m swaths that are actually continuous.


jessemapel:
go appveyeur

jessemapel:
In #260

PR #268, opened 2020-01-21
Added check for downtrack line direction

### LINKED ISSUE #267: New ground to image needs to determine if +lines on the detector is +lines in the image
For some images, we're seeing that +lines on the detector is -lines in the image, this causes groundtoimage to fail quite spectacularly. This can be accomplished with a simple check at initialization.

jessemapel:
Fixes #267 
Do not merge until GXP testing is finished.


jessemapel:
This is now passing on GXP, I had forgotten to add the new member variable to the state

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -609,6 +615,22 @@ void UsgsAstroLsSensorModel::updateState()
    MESSAGE_LOG(m_logger, "updateState: half time duration set to {}",
                                m_halfTime)
 
+   // Compute if downtrack is increasing or decreasing lines
+   // This requires that all of the other parameters be set first!
+   csm::ImageCoord nextLinePoint((lineCtr+m_nLines)/2, sampCtr);
```
acpaquette: I would also mention why we are even worried about the line track. Potentially mention the LROC and KaguyaTC problems we were/are having.

include/usgscsm/UsgsAstroLsSensorModel.h
```diff
@@ -111,6 +111,7 @@ class UsgsAstroLsSensorModel : public csm::RasterGM, virtual public csm::Settabl
    std::vector<double> m_currentParameterValue;
    std::vector<csm::param::Type> m_parameterType;
    csm::EcefCoord m_referencePointXyz;
+   int          m_downtrackLines;
```
acpaquette: Another member variable :(

PR #271, opened 2020-01-22
Implemented cross covariance method

### LINKED ISSUE #270: UsgsAstroLsSensorModel::getCrossCovarianceMatrix returns zeros
The function begins gathering the parameters and the just returns a vector (rows*columns large) of all zeros. To view this function look at UsgsAstroLsSensorModel.cpp line 1709.

jessemapel:
Fixes #270

### DIFF
--- a/src/UsgsAstroLsSensorModel.cpp
+++ b/src/UsgsAstroLsSensorModel.cpp
@@ -632,18 +632,37 @@ void UsgsAstroLsSensorModel::updateState()
    MESSAGE_LOG(m_logger, "updateState: down track line direction set to {}",
                                m_downtrackLines)
 
-   // Parameter covariance, hardcoded accuracy values
+   // Parameter covariance
+   // hardcoded ~1 pixel accuracy values
    int num_params = NUM_PARAMETERS;
-   int num_paramsSquare = num_params * num_params;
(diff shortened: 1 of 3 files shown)


src/UsgsAstroLsSensorModel.cpp
```diff
@@ -1940,7 +1971,8 @@ void UsgsAstroLsSensorModel::losToEcf(
   // Define imaging ray (look vector) in camera space
    double cameraLook[3];
    createCameraLookVector(undistortedFocalPlaneX, undistortedFocalPlaneY,
-                          m_zDirection, m_focalLength + getValue(15, adj), cameraLook);
+                          m_zDirection, m_focalLength * (1 - getValue(15, adj) / m_halfSwath),
```
jessemapel: This changes it back to the previous behavior. This also mean that adjusting the focal length adjustment results in the same adjustment as the other parameters.

jessemapel: https://github.com/USGS-Astrogeology/usgscsm/blame/eb02f11be6e035bb4c0ab276637e05fc4d419788/src/UsgsAstroLsSensorModel.cpp#L1739

PR #273, opened 2020-02-07
ImageToGround and groundToImage changes

acpaquette:
Updates imageToGround to use values read from the isd and converts the detector line to image line


jessemapel:
Latest change has fixed imageToGround and groundToImage not inverting properly, but tests still need to be updated as the losToEcf change has shifted us by about a half pixel.

jessemapel:
Changes have been verified by running real data tests. @acpaquette this can be merged whenever.

PR #274, opened 2020-02-07
Revert "Added check for downtrack line direction"

acpaquette:
Reverts USGS-Astrogeology/usgscsm#268


jessemapel:
For future reference, we're addressing this at the ISD level instead of the sensor mode level

PR #275, opened 2020-04-09
ALE Isd Object Update

acpaquette:
This PR updates usgscsm to use the ALE isd object


src/UsgsAstroLsSensorModel.cpp
```diff
@@ -2877,7 +2923,7 @@ std::string UsgsAstroLsSensorModel::constructStateFromIsd(const std::string imag
   }
 
   // Get the optional logging file
-  state["m_logFile"] = getLogFile(isd);
+  state["m_logFile"] = "";
```
kberryUSGS, line 2926: Temporary, to be re-added later?

jessemapel:
This is now passing tests locally, but the comparisons in Knoten are not passing yet. So, ALE likely has an issue somewhere and before that gets a new release published we should fix it. I am going to leave this up as a draft and then once the ALE release happens this can be merged in.

jessemapel:
I am merging this into the branch so that we can keep it in sync.

PR #276, opened 2020-04-13
Removed old O-REx model

jessemapel:
This sensor model was never finished and we have no plans to re-visit so it is being removed. If we add support for OCAMS, then it will be through the generic framing sensor model.


jessemapel:
travis Mac is having weird intermittent failures.

jlaura:
I relaunched once before approving seeing it was Mac having Travis issues. Perhaps just rerun in a few hours? Time seems to be Travis' best medicine.

PR #277, opened 2020-04-14
Added root finding algorithm

jessemapel:
This adds Brent's root finding algorithm which we will need to compute the 0 doppler shift time and the ground range in the SAR groundToImage method.

### DIFF
--- a/include/usgscsm/Utilities.h
+++ b/include/usgscsm/Utilities.h
@@ -69,6 +69,14 @@ void lagrangeInterp (
   const int&     i_order,
   double*        valueVector);
 
+// Brent's algorithm for finding the roots of a function
+// Arguments are two inputs that bracket a root, the function, and a convergence tolerance
+double brentRoot(
+  double lowerBound,
+  double upperBound,
+  double (*func)(double),
+  double epsilon = 1e-10);
+
 // Methods for checking/accessing the ISD
 
 double metric_conversion(double val, std::string from, std::string to="m");
--- a/src/Utilities.cpp
+++ b/src/Utilities.cpp
@@ -4,6 +4,7 @@
 #include <Error.h>
 #include <stack>
 #include <utility>
+#include <stdexcept>
 
 using json = nlohmann::json;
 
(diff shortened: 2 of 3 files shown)


jessemapel:
[Implement Brent's method for root finding](https://trello.com/c/JXpn3m3Q/285-implement-brents-method-for-root-finding)

include/usgscsm/Utilities.h
```diff
@@ -69,6 +69,14 @@ void lagrangeInterp (
   const int&     i_order,
   double*        valueVector);
 
+// Brent's algorithm for finding the roots of a function
+// Arguments are two inputs that bracket a root, the function, and a convergence tolerance
+double brentRoot(
```
thareUSGS, line 74: Keeping comment is fine, but do we really want the function named after someone? :-)

jessemapel: It's the name of the algorithm, so I don't see a reason to change it.

src/Utilities.cpp
```diff
@@ -276,6 +277,74 @@ void lagrangeInterp(
...
+        nextPoint += (currentPoint * counterFunc * previousFunc) / ((currentFunc - counterFunc) * (currentFunc - previousFunc));
+        nextPoint += (previousPoint * currentFunc * counterFunc) / ((previousFunc - counterFunc) * (previousFunc - currentFunc));
+      }
+      // Secant method
+      else {
+        nextPoint = currentPoint - currentFunc * (currentPoint - counterPoint) / (currentFunc - counterFunc);
+      }
+
+      // Bisection method
+      if (((currentPoint - nextPoint) * (nextPoint - (3 * counterPoint + currentPoint) / 4) < 0) ||
```
kberryUSGS, line 318: It would still be nice to have the comments you mentioned potentially adding in here about what the various if-checks in here are doing.

kberryUSGS: But they're also really clearly enumerated in https://en.wikipedia.org/wiki/Brent%27s_method, so no big deal if you don't think it's worth adding.

include/usgscsm/Utilities.h
thareUSGS: haha - I thought it was our Brent and he was simply pointing to a generic root finding method.

kberryUSGS: @thareUSGS At first, I thought that, toO!

src/Utilities.cpp
jessemapel: I think I'm just going to leave the comment out. If explanation if needed it's better to just point people to either a lecture or wikipedia.

PR #279, opened 2020-04-16
Add SAR ISD to State method

jessemapel:
The test ISD should be good for actual testing, it's an orbital observation with the same line and sample width.


kberryUSGS:
Thanks for also fixing the failing Travis tests!

PR #280, opened 2020-04-20
Adds ground to image and support functions.

kberryUSGS:
Notable: 

* I know secantRoot needs to be cleaned up and probably moved to Utilities if it's useful. For some reason, I thought it was a good idea to pull this out into a function that takes a lambda about an hour and a half ago. Feedback still useful and requested!

* Looking for your opinion on the best approach for defining functions which need root finding. Ran into a problem where the functions I want to find the roots for _mathematically_ are of one variable, x, but have additional input parameters used as constants in the equation. These values either need to be passed in as input parameters,  _or_ put into the state of the object so they can just be accessed, _or_ instead we can define the function as a lambda inside another function and use the closure to capture the constant values needed. I put a (rough) example of this in the function that calls `secantRoot` as an example for discussion. I'm not sure I like this, but it's the only way that made consistent sense to me. Let me know what you think. 

* You'll also notice a lot of "how to fail"? messages in here. Is there a standard way in CSM to fail? 

* Also, there are a lot of placeholders for "accessing values from the ISD" because I wasn't sure which of these you were working on. These obviously need to be updated.


src/UsgsAstroSarSensorModel.cpp
```diff
@@ -1,13 +1,18 @@
 #include "UsgsAstroSarSensorModel.h"
 #include "Utilities.h"
 
+#include <functional>
+#include <string.h>
```
jessemapel, line 5: String is already included

src/UsgsAstroSarSensorModel.cpp
```diff
@@ -136,3 +141,209 @@ string UsgsAstroSarSensorModel::constructStateFromIsd(
...
+  csm::EcefCoord newGroundPt(groundPt);
+  double time = dopplerShiftRoot(newGroundPt);
+  double slantRange = calcSlantRange(newGroundPt, time);
+
+  if (time != -1) {
+    // Find the ground range, based on the ground-range-to-slant-range polynomial defined by the 
+    // range coefficient set, with a time closest to the calculated time of closest approach
+    double groundRange = slantToGroundRange(groundPt, time, slantRange);
+
+    if (groundRange =! -1) {
```
jessemapel, line 185: comparing a float to -1 seems dubious here.

src/UsgsAstroSarSensorModel.cpp
```diff
@@ -136,3 +141,209 @@ string UsgsAstroSarSensorModel::constructStateFromIsd(
...
+      double line = 1 + (time - START_TIME) / EXPOSURE_DURATION;
+      double sample = 1 + groundRange / SCALED_PIXEL_WIDTH;
+      return csm::ImageCoord(line, sample);
+    }
+  }
+  return csm::ImageCoord(-1, -1); // what do we do if this fails?
```
jessemapel, line 191: If it completely fails we should raise an exception. See Error.h in CSM

src/UsgsAstroSarSensorModel.cpp
```diff
@@ -136,3 +141,209 @@ string UsgsAstroSarSensorModel::constructStateFromIsd(
...
+  double slantRange = calcSlantRange(newGroundPt, time);
+
+  if (time != -1) {
+    // Find the ground range, based on the ground-range-to-slant-range polynomial defined by the 
+    // range coefficient set, with a time closest to the calculated time of closest approach
+    double groundRange = slantToGroundRange(groundPt, time, slantRange);
+
+    if (groundRange =! -1) {
+      double line = 1 + (time - START_TIME) / EXPOSURE_DURATION;
+      double sample = 1 + groundRange / SCALED_PIXEL_WIDTH;
```
jessemapel, line 187: Socet Set and CSM use different coordinate systems. I think the 1+ needs to go away in each of these.

src/UsgsAstroSarSensorModel.cpp
```diff
@@ -136,3 +141,209 @@ string UsgsAstroSarSensorModel::constructStateFromIsd(
...
+
+     lookVector[0] = spacecraftPosition[0] - surfPt.x;
+     lookVector[1] = spacecraftPosition[1] - surfPt.y;
+     lookVector[2] = spacecraftPosition[2] - surfPt.z;
+
+     // Existing vector magnitude function in usgscsm? 
+     double slantRange = sqrt(pow(lookVector[0], 2) +  pow(lookVector[1], 2) + pow(lookVector[2], 2)); 
+
+
+     // Existing dot product function in usgscsm?
```
jessemapel, line 215: None of these functions are in CSM

src/UsgsAstroSarSensorModel.cpp
```diff
@@ -136,3 +141,209 @@ string UsgsAstroSarSensorModel::constructStateFromIsd(
...
+
+  // Lower-bound for doppler-shift
+  double startShift = dopplerShiftFrequency(START_TIME);
+
+  // Upper-bound for doppler-shift
+  double stopShift = dopplerShiftFrequency(STOP_TIME);
+
+  // Make sure we bound root (dopplerShift = 0.0)
+  if (((startShift < 0.0) && (stopShift < 0.0)) || ((startShift > 0.0) && (stopShift > 0.0))) {
+    return -1;
```
jessemapel, line 229: Instead of doing C-style return code, this should just raise an exception that gets caught higher up.

src/UsgsAstroSarSensorModel.cpp
```diff
@@ -136,3 +141,209 @@ string UsgsAstroSarSensorModel::constructStateFromIsd(
...
+      }
+      else {
+        deltaTime = xh - time;
+        xh = time;
+        fh = shift;
+      }
+
+      // See if we are done
+      if ((fabs(deltaTime) <= tolerance) || (shift == 0.0)) {
+        time = time;
```
jessemapel, line 281: time = time?

src/UsgsAstroSarSensorModel.cpp
```diff
@@ -136,3 +141,209 @@ string UsgsAstroSarSensorModel::constructStateFromIsd(
...
+
+  return secantRoot(dopplerShiftFrequency, xl, xh, tolerance, 30);
+}
+
+
+double secantRoot(std::function<double(double)> lambda, double xl, double xh, 
+                  double tolerance, double maxIters) {
+  double time;
+  int j = 0;
+    int found = -1;
```
jessemapel, line 255: Found should just be a boolean.

src/UsgsAstroSarSensorModel.cpp
```diff
@@ -136,3 +141,209 @@ string UsgsAstroSarSensorModel::constructStateFromIsd(
...
+        time = time;
+        found = 0;
+      }
+      j++;
+    }
+    if (found != -1) {
+      return time; 
+    }
+    else {
+      return -1;
```
jessemapel, line 290: This should raise an exception.

src/UsgsAstroSarSensorModel.cpp
```diff
@@ -136,3 +141,209 @@ string UsgsAstroSarSensorModel::constructStateFromIsd(
...
+  // If the ground range guesses at the 2 extremes of the image are equal
+  // or they have the same sign, then the ground range cannot be solved for.
+  if (!((minGroundRangeGuess == maxGroundRangeGuess) || 
+      (minGroundRangeGuess < 0.0 && maxGroundRangeGuess < 0.0) ||
+      (minGroundRangeGuess > 0.0 && maxGroundRangeGuess > 0.0))) {
+
+    // Use Brent's algorithm to find a root of the function:
+    return brentRoot(minGroundRangeGuess, maxGroundRangeGuess, groundRangeFunc, 0.1);
+  }
+  return -1; // How to fail?
```
jessemapel, line 342: Another exception here.

src/UsgsAstroSarSensorModel.cpp
kberryUSGS: You're right, thanks for catching this.

kberryUSGS: Are ISIS and CSM 1 off or 0.5 off?

src/UsgsAstroSarSensorModel.cpp
kberryUSGS: merge error, thanks

src/UsgsAstroSarSensorModel.cpp
kberryUSGS: Cleaning this up, not sure what that was supposed to be

src/UsgsAstroSarSensorModel.cpp
```diff
@@ -136,3 +140,149 @@ string UsgsAstroSarSensorModel::constructStateFromIsd(
...
+  double groundRange = slantToGroundRange(groundPt, time, slantRange);
+
+  double line = (time - START_TIME) / EXPOSURE_DURATION;
+  double sample = groundRange / SCALED_PIXEL_WIDTH;
+  return csm::ImageCoord(line, sample);
+
+  throw csm::Error(
+    csm::Error::UNKNOWN_ERROR,
+    "Could not calculate groundToImage.",
+    "groundToImage");
```
jessemapel, line 179: Why is this here? This will never run because of the return call.

src/UsgsAstroSarSensorModel.cpp
```diff
@@ -136,3 +140,149 @@ string UsgsAstroSarSensorModel::constructStateFromIsd(
...
+   };
+
+  // Lower-bound for doppler-shift
+  double startShift = dopplerShiftFrequency(START_TIME);
+
+  // Upper-bound for doppler-shift
+  double stopShift = dopplerShiftFrequency(STOP_TIME);
+
+  // Make sure we bound root (dopplerShift = 0.0)
+  if (!(((startShift < 0.0) && (stopShift < 0.0)) || ((startShift > 0.0) && (stopShift > 0.0)))) {
```
jessemapel, line 213: It's easier to check that startShift * stopShift < 0.0

kberryUSGS: Good point, thanks!

src/UsgsAstroSarSensorModel.cpp
kberryUSGS: Hey, thanks for catching this. I think this used to be in an 'else' but I've since moved the exceptions to the functions that that this function calls. I'm going to add a try/catch & rethrow this error in here.

src/UsgsAstroSarSensorModel.cpp
kberryUSGS: I just moved this check into the secantRoot function like you did for brentRoot, and removed this code.

kberryUSGS:
[Update ALE isd generation and all code in usgscsm that handles the range coefficients to split the values into a "times" list and a "coefficients" list](https://trello.com/c/iE12A3VD/349-update-ale-isd-generation-and-all-code-in-usgscsm-that-handles-the-range-coefficients-to-split-the-values-into-a-times-list-and)

PR #283, opened 2020-04-27
Added the look direction to SAR sensor model

kberryUSGS:
I just made this a `std::string` rather than an enum with left/right.


include/usgscsm/UsgsAstroSarSensorModel.h
```diff
@@ -249,6 +249,7 @@ class UsgsAstroSarSensorModel : public csm::RasterGM, virtual public csm::Settab
     std::vector<double> m_sunPosition;
     std::vector<double> m_sunVelocity;
     double m_wavelength;
+    std::string m_lookDirection;
```
jessemapel, line 252: I would rather look direction be an enumeration or something. Comparing a string to determine left/right is not good. It's better to have either an enumeration or maybe a boolean.

kberryUSGS:
[Update ale driver for radar to add left/right look direction (and usgscsm to read in)](https://trello.com/c/jILhGvo9/353-update-ale-driver-for-radar-to-add-left-right-look-direction-and-usgscsm-to-read-in)

include/usgscsm/UsgsAstroSarSensorModel.h
kberryUSGS: Yeah, you're right. I'd thought of making it an enum, and I guess I just got lazy.

PR #285, opened 2020-04-28
More stubs

jessemapel:
Added a bunch more methods. Most of this is copy and paste from the line scanner.

### DIFF
--- a/src/UsgsAstroSarSensorModel.cpp
+++ b/src/UsgsAstroSarSensorModel.cpp
@@ -627,32 +627,47 @@ pair<double, double> UsgsAstroSarSensorModel::getValidHeightRange() const
 
 csm::EcefVector UsgsAstroSarSensorModel::getIlluminationDirection(const csm::EcefCoord& groundPt) const
 {
-  return csm::EcefVector(0.0, 0.0, 0.0);
+  csm::EcefVector sunPosition = getSunPosition(getImageTime(groundToImage(groundPt)));
+  csm::EcefVector illuminationDirection = csm::EcefVector(groundPt.x - sunPosition.x,
+                                                          groundPt.y - sunPosition.y,
+                                                          groundPt.z - sunPosition.z);
+
+  double scale = sqrt(illuminationDirection.x * illuminationDirection.x +
+                      illuminationDirection.y * illuminationDirection.y +
+                      illuminationDirection.z * illuminationDirection.z);
+
+  illuminationDirection.x /= scale;
+  illuminationDirection.y /= scale;
+  illuminationDirection.z /= scale;
+  return illuminationDirection;
 }
 
 double UsgsAstroSarSensorModel::getImageTime(const csm::ImageCoord& imagePt) const
 {
-  return 0.0;
+  return m_startingEphemerisTime + (imagePt.line - 0.5) * m_exposureDuration;
 }
 
 csm::EcefCoord UsgsAstroSarSensorModel::getSensorPosition(const csm::ImageCoord& imagePt) const
 {
-  return csm::EcefCoord(0.0, 0.0, 0.0);
+  double time = getImageTime(imagePt);
+  return getSensorPosition(time);
 }
 
 csm::EcefCoord UsgsAstroSarSensorModel::getSensorPosition(double time) const
 {
-  return csm::EcefCoord(0.0, 0.0, 0.0);
+  csm::EcefVector sensorVector = getSpacecraftPosition(time);
+  return csm::EcefCoord(sensorVector.x, sensorVector.y, sensorVector.z);
 }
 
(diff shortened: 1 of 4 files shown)


src/UsgsAstroSarSensorModel.cpp
```diff
@@ -427,9 +427,64 @@ csm::ImageCoordCovar UsgsAstroSarSensorModel::groundToImage(
     double* achievedPrecision,
     csm::WarningList* warnings) const
 {
-  return csm::ImageCoordCovar(0.0, 0.0,
-                              1.0, 0.0,
-                                   1.0);
+  // Ground to image with error propagation
+  // Compute corresponding image point
+  csm::EcefCoord gp;
```
kberryUSGS, line 432: Does `gp(groundPt)` not do the right thing? Or `gp(groundPt.x, groundPt.y, groundPt.z)`. I like these two better than separating out into the next few lines.

src/UsgsAstroSarSensorModel.cpp
```diff
@@ -513,47 +635,64 @@ csm::ImageCoord UsgsAstroSarSensorModel::getImageStart() const
...
-  return csm::ImageVector(0.0, 0.0);
+  return csm::ImageVector(m_nLines, m_nSamples);
 }
 
 pair<csm::ImageCoord, csm::ImageCoord> UsgsAstroSarSensorModel::getValidImageRange() const
 {
-  return make_pair(csm::ImageCoord(0.0, 0.0), csm::ImageCoord(0.0, 0.0));
+  csm::ImageCoord start = getImageStart();
+  csm::ImageVector size = getImageSize();
+  return make_pair(start, csm::ImageCoord(start.line + size.line, start.samp + start.samp));
```
kberryUSGS, line 645: Should the last one be `start.samp + size.samp` ?

src/UsgsAstroSarSensorModel.cpp
```diff
@@ -513,47 +635,64 @@ csm::ImageCoord UsgsAstroSarSensorModel::getImageStart() const
...
 {
-  return make_pair(0.0, 0.0);
+  return make_pair(m_minElevation, m_maxElevation);
 }
 
 csm::EcefVector UsgsAstroSarSensorModel::getIlluminationDirection(const csm::EcefCoord& groundPt) const
 {
-  return csm::EcefVector(0.0, 0.0, 0.0);
+  csm::EcefVector sunPosition = getSunPosition(getImageTime(groundToImage(groundPt)));
+  csm::EcefVector illuminationDirection = csm::EcefVector(groundPt.x - sunPosition.x,
```
kberryUSGS, line 656: Might as well use the utility functions you just added to do vector operations? 

like `illuminationDirection = groundPt - sunPosition`?  But then I guess you'd need to make a vector for groundPt. Not sure whether it's worth it or not.

kberryUSGS: And norm(illuminationDirection) below.

src/UsgsAstroSarSensorModel.cpp
```diff
@@ -513,47 +635,64 @@ csm::ImageCoord UsgsAstroSarSensorModel::getImageStart() const
...
 {
-  return csm::EcefVector(0.0, 0.0, 0.0);
+  double time = getImageTime(imagePt);
+  return getSensorVelocity(time);
 }
 
 csm::EcefVector UsgsAstroSarSensorModel::getSensorVelocity(double time) const
 {
-  return csm::EcefVector(0.0, 0.0, 0.0);
+  return getSpacecraftVelocity(time);
```
kberryUSGS, line 695: Is the separate `getSpacecraftVelocity` function required by the API, or can we just delete it and replace it with `getSensorVelocity`?

src/UsgsAstroSarSensorModel.cpp
jessemapel: Yeah, we should actually be able to just pass the EcefCoordCovar into the functions and it will get sliced but that's okay. I think the whole making a copy part can just go away.

src/UsgsAstroSarSensorModel.cpp
jessemapel: Yeah all of this was written before I added those. I need to go back over it and use those!

src/UsgsAstroSarSensorModel.cpp
jessemapel: The getSpacecraft* functions are not part of the API and I think we can just axe them in favor of the getSensor* functions which are part of the API.

src/UsgsAstroSarSensorModel.cpp
jessemapel: so, we do have to make a new ground point for const reasons, but your suggestion of using the copy constructor works.

PR #286, opened 2020-04-29
Adds SAR imaging loci functions and fixes imagetoground

jessemapel:
Added both imaging loci functions, ellipsoid intersection in image to ground, and look direction in image to ground.

### DIFF
--- a/src/UsgsAstroSarSensorModel.cpp
+++ b/src/UsgsAstroSarSensorModel.cpp
@@ -526,7 +526,9 @@ csm::EcefCoord UsgsAstroSarSensorModel::imageToGround(
       beta *= -1;
     }
     groundVec = alpha * tHat + beta * uHat + spacecraftPosition;
-    computeElevation(groundVec.x, groundVec.y, groundVec.z, pointHeight);
+    pointHeight = computeEllipsoidElevation(
+        groundVec.x, groundVec.y, groundVec.z,
+        m_majorAxis, m_minorAxis);
     pointRadius -= (pointHeight - height);
   } while(fabs(pointHeight - height) > desiredPrecision);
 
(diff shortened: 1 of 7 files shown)


src/UsgsAstroSarSensorModel.cpp
```diff
@@ -1074,3 +1136,64 @@ csm::EcefVector UsgsAstroSarSensorModel::getSunPosition(const double imageTime)
...
+         cotPhi = dd / z;
+         ktr++;
+      } while (MKTR > ktr && fabs(h - hPrev) > desired_precision);
+   }
+
+   height = h;
+   if (achieved_precision) {
+     *achieved_precision = fabs(h - hPrev);
+   }
+}
```
jessemapel, line 1199: Copy and pasted from LS. Should this be moved into utilities and have the radii passed in?

tests/SarTests.cpp
```diff
@@ -48,12 +48,12 @@ TEST_F(SarSensorModel, Center) {
   csm::EcefCoord groundPt = sensorModel->imageToGround(imagePt, 0.0);
   // TODO these tolerances are bad
   EXPECT_NEAR(groundPt.x, 1737391.90602155, 1e-2);
-  EXPECT_NEAR(groundPt.y, 3749.98835331, 1e-2);
+  EXPECT_NEAR(groundPt.y, -3749.98835331, 1e-2);
```
jessemapel, line 51: These negations are because the sensor was hard-coded to left looks and the test image is a right look.

src/UsgsAstroSarSensorModel.cpp
kberryUSGS: Yeah, I think it's worth moving this into Utilities and passing in the radii.

tests/SarTests.cpp
kberryUSGS: Is the hard-coding to a left look something that will get fixed in a future PR?

jessemapel:
[Image2Ground on an ellipsoid.](https://trello.com/c/qxwSPfaO/360-image2ground-on-an-ellipsoid)

tests/SarTests.cpp
kberryUSGS: It did absolutely look like the left vs. right look was being handled, so just an old comment makes more sense!

PR #288, opened 2020-05-05
Added adjustable parameters.

kberryUSGS:


src/UsgsAstroSarSensorModel.cpp
```diff
@@ -355,6 +355,42 @@ csm::ImageCoord UsgsAstroSarSensorModel::groundToImage(
...
+    csm::WarningList* warnings) const {
+
+  // Find time of closest approach to groundPt and the corresponding slant range by finding
+  // the root of the doppler shift frequency
+  try {
+    double timeTolerance = m_exposureDuration * desiredPrecision / 2.0;
+
+    // Adjust the ground point value (I think. Check notes... maybe not) 
+    csm::EcefCoord adjGroundPt(groundPt.x + getValue(0, adj), groundPt.y + getValue(1, adj), 
+                                   groundPt.z+ getValue(2, adj));
```
jessemapel, line 372: These adjustments need to be added to the sensor position, not the ground point.

src/UsgsAstroSarSensorModel.cpp
```diff
@@ -355,6 +355,42 @@ csm::ImageCoord UsgsAstroSarSensorModel::groundToImage(
...
+    double time = dopplerShift(adjGroundPt, timeTolerance);
+    double slantRangeValue = slantRange(groundPt, time);
```
jessemapel, line 374: The adjustment vector should be passed to these functions because they get the sensor position and velocity.

src/UsgsAstroSarSensorModel.cpp
```diff
@@ -825,9 +926,41 @@ vector<csm::RasterGM::SensorPartials> UsgsAstroSarSensorModel::computeAllSensorP
...
+   size_t num = indices.size();
+   std::vector<csm::RasterGM::SensorPartials> partials;
+   for (int index = 0; index < num; index++)
+   {
+     partials.push_back(
+         computeSensorPartials(
+            indices[index], imagePt, groundPt,
+            desiredPrecision, achievedPrecision, warnings));
+   }
+   return partials;
```
jessemapel, line 960: We don't need this function, it's already in the RasterGM class from the CSM API.

tests/SarTests.cpp
```diff
@@ -114,3 +114,16 @@ TEST_F(SarSensorModel, imageToRemoteImagingLocus) {
...
+  std::vector<double> adjustments = {1000000.0, 0.2, -10.0, -20.0, 0.5, 2000000.0};
+  csm::EcefCoord adjPosition = sensorModel->getAdjustedSensorPosition(0.0, adjustments);
+  csm::EcefVector adjVelocity = sensorModel->getAdjustedSensorVelocity(0.0, adjustments);
+  
+  EXPECT_NEAR(adjPosition.x, 4737399.9991294323, 1e-2);
+  EXPECT_NEAR(adjPosition.y, 0.2, 1e-2);
+  EXPECT_NEAR(adjPosition.z, -18.0667951486008143, 1e-2);
+  EXPECT_NEAR(adjVelocity.x, -28.0667951486008143, 1e-8);
+  EXPECT_NEAR(adjVelocity.y, 0.5, 1e-8);
+  EXPECT_NEAR(adjVelocity.z, -1737399.9991294323, 1e-2);
```
jessemapel, line 128: Don't hard code the values here. Use the un-adjusted position and velocity to get the comparison values.

kberryUSGS: Can do. That's what I did by-hand and you're right that it makes more sense to just do it in here.

src/UsgsAstroSarSensorModel.cpp
kberryUSGS: Does this make sense to pull out of linescan, then, too? (future ticket?)

jessemapel: Yeah that sounds good as a future ticket.

Issue #255, opened 2020-01-07
Convert logging to be based on an environment variable


jessemapel:
The current logging setup uses a value passed through the ISD. This presents two problems:

1. We cannot log in the plugin
1. We cannot differentiate between multiple sensor models created from the same ISD

We should consider allowing users to set an environment variable that enables logging and specifies a logging directory. Then, we can have a single log file for the plugin and each sensor model can have its own log file based on its image identifier and some additional unique id.

PR #289, opened 2020-06-24
Logging Update

acpaquette:
Changes how logging is handled in both the Framer and Line Scanner sensor models. The plugin now handles logger creation and each sensor model will only log if this logger is present.

### Detail

Whenever a plugin is in instantiated, the plugin will also create a logger called `usgscsm_logger` which will have a sink defined in the environment variable `ALE_LOG_FILE`. The two sensor models write to this logger by default.

If you want to adjust where either of the sensor models write to, you'll need to create a new spd logger, register said logger, and update the name of the logger for the sensor model you are working with. As such this can only really be done on the C++ side but the output location of your log can be changed on the fly by changing `ALE_LOG_FILE`.

A few things to potentially discuss:

Do we want to be able to split the loggers more dynamically rather than everything be dumped into one output location?

Do we just create a config that has a separate logger and log output location that can be parsed at run time for each sensor model?

### DIFF
--- a/src/UsgsAstroFrameSensorModel.cpp
+++ b/src/UsgsAstroFrameSensorModel.cpp
@@ -874,7 +874,7 @@ std::string UsgsAstroFrameSensorModel::constructStateFromIsd(const std::string& 
     }
 
     MESSAGE_LOG("Constructing state from isd");
-    ale::Isd stateIsd(jsonIsd);
+    json isd = json::parse(jsonIsd);
 
     csm::WarningList* parsingWarnings = new csm::WarningList;
 
--- a/src/UsgsAstroFrameSensorModel.cpp
+++ b/src/UsgsAstroFrameSensorModel.cpp
@@ -9,7 +9,7 @@
 #include <Error.h>
 #include <Version.h>
 
-#define MESSAGE_LOG(logger, ...) if (logger) { logger->info(__VA_ARGS__); }
+#define MESSAGE_LOG(...) if (spdlog::get(m_logName)) { spdlog::get(m_logName)->info(__VA_ARGS__); }
 
 using json = nlohmann::json;
 using namespace std;
@@ -77,8 +77,7 @@ void UsgsAstroFrameSensorModel::reset() {
     m_referencePointXyz.x = 0;
     m_referencePointXyz.y = 0;
     m_referencePointXyz.z = 0;
-    m_logFile = "";
-    m_logger.reset();
+    m_logName = "usgscsm_logger";
 }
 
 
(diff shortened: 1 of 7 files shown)


src/UsgsAstroFrameSensorModel.cpp
```diff
@@ -875,9 +865,16 @@ void UsgsAstroFrameSensorModel::replaceModelState(const std::string& stringState
...
 
 std::string UsgsAstroFrameSensorModel::constructStateFromIsd(const std::string& jsonIsd, csm::WarningList* warnings) {
-    MESSAGE_LOG(this->m_logger, "Constructing state from isd");
-    json isd = json::parse(jsonIsd);
+
     json state = {};
+    // Get the optional custom logger
+    state["m_logName"] = "usgscsm_logger";
+    if (m_logName != "usgscsm_logger" && m_logName != "") {
+      state["m_logName"] = m_logName;
```
jessemapel, line 873: Why is this inserting the model's existing log name?

src/UsgsAstroPlugin.cpp
```diff
@@ -36,6 +36,18 @@ const int         UsgsAstroPlugin::_N_SENSOR_MODELS = 3;
 const UsgsAstroPlugin UsgsAstroPlugin::m_registeredPlugin;
 
 UsgsAstroPlugin::UsgsAstroPlugin() {
+
+  // Build and register the USGSCSM logger on pluggin creation
+  std::string logFile = getenv("ALE_LOG_FILE");
+  if (logFile != "") {
+    std::shared_ptr<spdlog::logger> logger = spdlog::get("usgscsm_logger");
+    if (!logger) {
+      logger = spdlog::basic_logger_mt("usgscsm_logger", logFile);
```
jessemapel, line 45: I think this is missing a make_shared call.

src/UsgsAstroPlugin.cpp
```diff
@@ -36,6 +36,18 @@ const int         UsgsAstroPlugin::_N_SENSOR_MODELS = 3;
...
+      try {
+        spdlog::register_logger(logger);
+      } catch (...) {}
```
jessemapel, line 48: Why try catch here?

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -265,16 +265,7 @@ void UsgsAstroLsSensorModel::replaceModelState(const std::string &stateString )
...
-   if (m_logFile.empty()) {
-     m_logger.reset();
-   }
-   else {
-     m_logger = spdlog::get(m_logFile);
-     if (!m_logger) {
-       m_logger = spdlog::basic_logger_mt(m_logFile, m_logFile);
-     }
-   }
+   m_logName = j["m_logName"].get<std::string>();
```
jessemapel, line 268: I don't think we actually need the logName to be a member variable or a part of the state at all. There's no way for users to setup a new logger, so why would they specify a new logger name?

acpaquette:
Discussed above comments from @jessemapel and decided that the sensor models could have a pointer to the logger instead of carrying around the logger name and getting the logger every time to log a message. Also resolved the try catch blocks around the register commands for the loggers.

jessemapel:
Fixes #255

PR #290, opened 2020-06-26
Fixed quaternion order and times

jessemapel:
With this fix, LROC is now matching what it used to 🎉


jessemapel:
MEX HRSC, Kaguya TC, and MRO CTX are also back where they were. So, this gets our LS back to parity and uses the new ISD stuff.

jessemapel:
I still need to update tests on this, but right now I am working on getting the framer matching.

jessemapel:
Framers are now matching too. This required some additional ALE changes because of the ISD not working great for the framer. This is ready for review and then in a new PR we can remove the ISD object and just parse the JSON instead.

src/UsgsAstroFrameSensorModel.cpp
```diff
@@ -435,14 +437,14 @@ csm::EcefVector UsgsAstroFrameSensorModel::getSensorVelocity(const csm::ImageCoo
 
 csm::EcefVector UsgsAstroFrameSensorModel::getSensorVelocity(double time) const {
```
kberryUSGS, line 438: If it's not used, is it still necessary to pass in `time`? Or is this part of the CSM api?

src/UsgsAstroFrameSensorModel.cpp
```diff
@@ -966,15 +966,17 @@ std::string UsgsAstroFrameSensorModel::constructStateFromIsd(const std::string&
...
 
-    for (ale::Rotation rotation : sensor_to_target.getRotations()) {
+    for (int i = 0; i < ephemTime.size(); i++) {
+      ale::Rotation rotation = sensor_to_target.interpolate(ephemTime[i], ale::SLERP);
       quaternion = rotation.toQuaternion();
-      quaternions.push_back(quaternion[0]);
       quaternions.push_back(quaternion[1]);
       quaternions.push_back(quaternion[2]);
       quaternions.push_back(quaternion[3]);
+      quaternions.push_back(quaternion[0]);
```
kberryUSGS, line 979: :+1: I don't know what this was before, so did we go from w, x, y, z to x, y, z, w or the other way around?

src/UsgsAstroFrameSensorModel.cpp
jessemapel: This is part of the API

src/UsgsAstroFrameSensorModel.cpp
jessemapel: The USGSCSM rotation code uses W last notation and ALE uses W first notation.

kberryUSGS: @jessemapel Thanks!

PR #292, opened 2020-07-02
Update to use new ALE ISD format

jessemapel:
This PR can't be merged until we tag a new ALE release.


jessemapel:
ALe 0.8.2 is now out, so this can be re-opened

jessemapel:
Still getting ALE 0.8.1 :(

jessemapel:
Linking errors x.x

tests/data/orbitalFramer.json
```diff
@@ -31,29 +32,61 @@
...
+    "ephemeris_times": [
+      50
+    ],
+		"reference_frame": 1
+  },
+  "instrument_pointing": {
+    "quaternions": [
+      [-0.707106781186547, 0.0, -0.707106781186547, 0]
+    ],
+		"angular_velocities": [
```
kberryUSGS, line 52: remove indent here to align?

tests/data/orbitalFramer.json
```diff
@@ -31,29 +32,61 @@
...
   },
   "starting_detector_line": 0,
   "starting_detector_sample": 0,
+  "detector_sample_summing": 1.0,
+  "detector_line_summing": 1.0,
+  "starting_ephemeris_time": 0,
   "sun_position": {
     "positions": [
-      [100.0, 100.0, 0.0]
+      [100000.0, 100000.0, 0.0]
```
kberryUSGS, line 81: All the other sun positions in this PR were updated from meters to km. Just want to double check that this one is correct?

src/UsgsAstroFrameSensorModel.cpp
```diff
@@ -900,68 +922,95 @@ std::string UsgsAstroFrameSensorModel::constructStateFromIsd(const std::string&
...
-    state["m_ccdCenter"][0] = getDetectorCenterLine(isd, parsingWarnings);
-    state["m_ccdCenter"][1] = getDetectorCenterSample(isd, parsingWarnings);
+    state["m_ccdCenter"][0] = ale::getDetectorCenterLine(parsedIsd);
+    state["m_ccdCenter"][1] = ale::getDetectorCenterSample(parsedIsd);
 
 
     // get radii
-    state["m_minorAxis"] = getSemiMinorRadius(isd, parsingWarnings);
-    state["m_majorAxis"] = getSemiMajorRadius(isd, parsingWarnings);
+    state["m_minorAxis"] = ale::getSemiMinorRadius(parsedIsd) * 1000;
```
kberryUSGS, line 999: add a quick comment about units that require the * 1000 (ex: in km, but state needs to be in meters)?

src/UsgsAstroFrameSensorModel.cpp
```diff
@@ -900,68 +922,95 @@ std::string UsgsAstroFrameSensorModel::constructStateFromIsd(const std::string&
...
+    ale::Orientations sensor_to_j2000 = j2000_to_sensor.inverse();
+    ale::Orientations sensor_to_target = j2000_to_target * sensor_to_j2000;
+    ephemTime = sensor_to_target.getTimes();
+    std::vector<double> quaternion;
+    std::vector<double> quaternions;
+
+    for (int i = 0; i < ephemTime.size(); i++) {
+      ale::Rotation rotation = sensor_to_target.interpolate(ephemTime[i], ale::SLERP);
+      quaternion = rotation.toQuaternion();
+      quaternions.push_back(quaternion[1]);
```
kberryUSGS, line 969: Naming is a little confusing. This is still just one quaternion, but moving from w-first to w-last format, right?

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -2598,20 +2599,20 @@ std::string UsgsAstroLsSensorModel::constructStateFromIsd(const std::string imag
...
               state["m_nLines"].dump(), state["m_nSamples"].dump())
 
-  state["m_iTransS"] = getFocal2PixelSamples(isd, parsingWarnings);
-  state["m_iTransL"] = getFocal2PixelLines(isd, parsingWarnings);
+  state["m_iTransS"] = ale::getFocal2PixelSamples(jsonIsd);
+  state["m_iTransL"] = ale::getFocal2PixelLines(jsonIsd);
   MESSAGE_LOG("m_iTransS: {} "
               "m_iTransL: {} ",
-              state["m_iTransS"].dump(), state["m_iTransL"].dump())
+              state["m_iTransS"].dump(), state["m_iTransS"].dump())
```
kberryUSGS, line 2615: Second `m_ITransS`, should be `m_iTransL`?

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -2634,9 +2635,27 @@ std::string UsgsAstroLsSensorModel::constructStateFromIsd(const std::string imag
...
+  std::vector<ale::State> sunStates = sunState.getStates();
+  std::vector<double> ephemTime = sunState.getTimes();
+  ale::Orientations j2000_to_target = ale::getBodyRotation(jsonIsd);
+  ale::State rotatedSunState;
+  std::vector<double> sunPositions = {};
+  std::vector<double> sunVelocities = {};
+
+  for (int i = 0; i < ephemTime.size(); i++) {
+    rotatedSunState = j2000_to_target.rotateStateAt(ephemTime[i], sunStates[i]);
+    sunPositions.push_back(rotatedSunState.position.x * 1000);
```
kberryUSGS, line 2649: maybe this is redundant, but I always appreciate comments like "ale stores in km, but usgscsm in meters" or whatever the case is here for the * 1000.

tests/data/orbitalFramer.json
jessemapel: This was changed to 100,000 because the old sun position put the sun literally inside of the planet. This puts it a more realistic, but still smaller than it would be for the earth, distance away. Because the CSM API doesn't do vectors to the sun, it just does direction vectors, the magnitude doesn't impact the function output.

src/UsgsAstroFrameSensorModel.cpp
```diff
@@ -396,15 +400,15 @@ csm::EcefCoord UsgsAstroFrameSensorModel::getSensorPosition(const csm::ImageCoor
...
+    if (time == 0.0){
         csm::EcefCoord sensorPosition;
         sensorPosition.x = m_currentParameterValue[0];
         sensorPosition.y = m_currentParameterValue[1];
         sensorPosition.z = m_currentParameterValue[2];
 
         return sensorPosition;
     } else {
-        std::string aMessage = "Valid image time is %d", m_ephemerisTime;
+        std::string aMessage = "Valid image time is 0.0";
```
acpaquette, line 411: Is this specific to framers since there should only ever be one time?

jessemapel: Yes this is specific to the framer

PR #297, opened 2020-07-15
Added more logging

kberryUSGS:


include/usgscsm/UsgsAstroSarSensorModel.h
```diff
@@ -23,9 +25,9 @@ class UsgsAstroSarSensorModel : public csm::RasterGM, virtual public csm::Settab
 
     virtual std::string getModelState() const;
 
-    static std::string constructStateFromIsd(const std::string imageSupportData, csm::WarningList *list);
+    std::string constructStateFromIsd(const std::string imageSupportData, csm::WarningList *list);
```
kberryUSGS, line 28: These were both not static in the Framer and LineScanner, and it removed the ability to test using m_logger. Thoughts?

tests/SarTests.cpp
```diff
@@ -15,25 +15,28 @@
...
-  catch(...) {
-    for (auto &warn: warnings) {
-      std::cerr << "Warning in " << warn.getFunction() << std::endl;
-      std::cerr << "  " << warn.getMessage() << std::endl;
-    }
-    FAIL() << "constructStateFromIsd errored";
-  }
-  EXPECT_TRUE(warnings.empty());
-}
+// UsgsAstroSarSensorModel is the only sensor model with this function static. Static throws off the
```
kberryUSGS, line 18: See this comment. Thoughts?

src/UsgsAstroPlugin.cpp
```diff
@@ -273,97 +293,111 @@ std::string UsgsAstroPlugin::convertISDToModelState(const csm::Isd &imageSupport
...
+  
+  MESSAGE_LOG("ISD String: {}", stringIsd);
+  if (modelName == UsgsAstroFrameSensorModel::_SENSOR_MODEL_NAME) {
+    UsgsAstroFrameSensorModel *model =  new UsgsAstroFrameSensorModel();
+    try {
+      MESSAGE_LOG("Trying to construct a UsgsAstroFrameSensorModel");
+      model->replaceModelState(model->constructStateFromIsd(stringIsd, warnings));
+      std::shared_ptr<spdlog::logger> logger = model->getLogger();
+      if (logger) {
+        logger->info("Constructed model: {}", modelName);
```
jessemapel, line 307: Why does this not use the logging macro?

tests/SarTests.cpp
jessemapel: This doesn't need to be static

src/UsgsAstroPlugin.cpp
kberryUSGS: Ah! This was the only part of the plugin that was previously logged before I added to the Macro. Good catch, thanks!

PR #302, opened 2020-07-21
Replaced secant root with Newton's method in SAR

jessemapel:
The slantRangeToGroundRange function finds the zero of a polynomial using a generic secant method. It's faster and doesn't require bounding the root if we instead use Newton's method.

### DIFF
--- a/src/Utilities.cpp
+++ b/src/Utilities.cpp
@@ -279,12 +279,10 @@ void lagrangeInterp(
   }
 }
 
-double brentRoot(
-  double lowerBound,
-  double upperBound,
-  std::function<double(double)> func,
-  double epsilon)
-{
+double brentRoot(double lowerBound,
+                 double upperBound,
+                 std::function<double(double)> func,
+                 double epsilon) {
   double counterPoint = lowerBound;
   double currentPoint = upperBound;
   double counterFunc = func(counterPoint);
@@ -348,10 +346,8 @@ double brentRoot(
   return nextPoint;
 }
 
-double evaluatePolynomial(
-  const std::vector<double> &coeffs,
-  double x)
-{
+double evaluatePolynomial(const std::vector<double> &coeffs,
+                          double x) {
   if (coeffs.empty()) {
     throw std::invalid_argument("Polynomial coeffs must be non-empty.");
   }
@@ -365,15 +361,14 @@ double evaluatePolynomial(
   return value;
 }
 
-double evaluatePolynomialDerivative(
-  const std::vector<double> &coeffs,
-  double x)
-{
+double evaluatePolynomialDerivative(const std::vector<double> &coeffs,
+                                    double x) {
   if (coeffs.empty()) {
     throw std::invalid_argument("Polynomial coeffs must be non-empty.");
   }
   int i = coeffs.size() - 1;
   double value = i * coeffs[i];
+  --i;
   for (; i > 0; --i) {
     value *= x;
     value += i * coeffs[i];
@@ -381,12 +376,10 @@ double evaluatePolynomialDerivative(
   return value;
 }
 
-double polynomialRoot(
-  const std::vector<double> &coeffs,
-  double guess,
-  double threshold,
-  int maxIterations)
-{
+double polynomialRoot(const std::vector<double> &coeffs,
+                      double guess,
+                      double threshold,
+                      int maxIterations) {
   double root = guess;
   double polyValue = evaluatePolynomial(coeffs, root);
   double polyDeriv = 0.0;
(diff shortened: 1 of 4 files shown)


src/Utilities.cpp
```diff
@@ -283,125 +283,125 @@ double brentRoot(
...
-        currentPoint = nextPoint;
-        currentFunc = nextFunc;
-      }
-      else {
-        counterPoint = nextPoint;
-        counterFunc = nextFunc;
-      }
-    } while (++iteration < 30 && fabs(counterPoint - currentPoint) > epsilon);
+  double epsilon)
+{
```
kberryUSGS, line 287: Do we want this curly-brace placement to be standard for usgscsm, and if so, how do you feel about me adding a ticket to update the framer code to use this standard?

src/Utilities.cpp
```diff
@@ -283,125 +283,125 @@ double brentRoot(
...
+        (bisected && fabs(currentPoint - previousPoint) < epsilon) ||
+        (!bisected && fabs(previousPoint - evenOlderPoint) < epsilon)) {
+      nextPoint = (currentPoint + counterPoint) / 2;
+      bisected = true;
+    }
+    else {
+      bisected = false;
+    }
 
-double secantRoot(double lowerBound, double upperBound, std::function<double(double)> func,
```
kberryUSGS, line 350: Is this something we'd ever want to use elsewhere, if not here?

tests/UtilitiesTests.cpp
```diff
@@ -350,13 +350,20 @@ TEST(UtilitiesTests, brentRoot) {
...
-  EXPECT_THROW(secantRoot(-1000.0, 1000.0, testPoly), csm::Error);
+TEST(UtilitiesTests, polynomialEval) {
+  std::vector<double> coeffs = {-12.0, 4.0, -3.0, 1.0};
+  EXPECT_DOUBLE_EQ(evaluatePolynomial(coeffs, -1.0), -20.0);
+  EXPECT_DOUBLE_EQ(evaluatePolynomial(coeffs,  0.0), -12.0);
+  EXPECT_DOUBLE_EQ(evaluatePolynomial(coeffs,  2.0),  -8.0);
+  EXPECT_DOUBLE_EQ(evaluatePolynomial(coeffs,  3.5),   8.125);
+  EXPECT_THROW(evaluatePolynomial(std::vector<double>(), 0.0), std::invalid_argument);
+}
+
```
kberryUSGS, line 361: Could you add a test for `evaluatePolynomialDerivative()`?

src/Utilities.cpp
```diff
@@ -283,125 +283,125 @@ double brentRoot(
...
-  if (found) {
-   return x2;
-  }
-  else {
-    throw csm::Error(
-    csm::Error::UNKNOWN_ERROR,
-    "Could not find a root of the function using the secant method",
-    "secantRoot");
-  }
+  throw std::invalid_argument("Root finder did not converge in given number of iterations");
```
kberryUSGS, line 404: advocate for adding a number of iterations to the error so that the iterations can be "upped" if desired.

src/Utilities.cpp
```diff
@@ -283,125 +283,125 @@ double brentRoot(
...
+      return root;
     }
-
-    // Check to see if we're done
-    if ((fabs(diff) <= epsilon) || (f2 == 0.0) ) {
-      found = true;
-      break;
+    polyDeriv = evaluatePolynomialDerivative(coeffs, root);
+    if (fabs(polyDeriv) < 1e-15) {
+      throw std::invalid_argument("Derivative at guess is too close to 0.");
```
kberryUSGS, line 399: For debugging, it might be good to have the guess's value in here.

src/Utilities.cpp
jessemapel: We don't really have a standard right now. I had run a linter over the code once, but never committed it. I'm not sure if I like this or now. Maybe I should just change these?

src/Utilities.cpp
jessemapel: No, Brent's is superior in the generic function space and Newton's is superior for polynomials.

src/Utilities.cpp
kberryUSGS: I don't know either. I was honestly just thinking that if we have hours left and get all the actual SAR work done, it might be a nice bonus to decide on one way or the other and then standardize the code. (If I have time, I volunteer!)

kberryUSGS: As far as I can tell, Ls and Sar use:
```
double func(int x)
{
}
```

and framer uses: 

```
double func(int x){
}
```
I'm happy to call this "beyond the scope of this PR" and talk about this later (if we do end up having time.)

src/Utilities.cpp
kberryUSGS: Sounds good to me!

Issue #304, opened 2020-07-22
Run a linter over the code / normalize coding standards


kberryUSGS:
Different numbers of spaces are used for indents and different curly-brace placements are used (among other things) throughout the code.

kberryUSGS:
Also, for exception error message variable names, we seem to alternate between msg, message, aMessage, etc... might as well use the same name.

kberryUSGS:
Still needed: say in the README or contributing that we use the google style guide.

oleg-alexandrov:
If linter checks are enabled, also a program should be provided to fix linter errors, such as clang-format-8, with a hook at commit time (the user installs the program separately and the git hook calls it). Or else developing code can be a very annoying affair, with sometimes hundreds of trivial things to fix manually for each commit.

PR #307, opened 2020-07-22
Updated to use Google style for formatting

kberryUSGS:
This PR contains the result of running just `clang-format style=Google` on the usgscsm code. This standardizes spacing, bracket-placement, header-ordering, max line length, etc, but doesn't contain changes resulting from static analysis and `cpplint` does still find some additional changes that need to be made by hand. (I'm not sure if these additional changes should be in _this_ PR or a subsequent one. I was leaning toward a subsequent one just so the automated changes and hand-changes could be in separate PRs. The changes left to be made are a much smaller list than the changes in this PR.)

I tested a few formats, and picked Google style because I somewhat arbitrarily liked the way the results looked the best. I have results locally for the following formats if anyone would like to see some alternatives: LLVM, GNU, Mozilla, Chromium, WebKit. I picked these to test because `clang-format` could automatically fix code to follow these standards. I'm also open to other "auto-fix formatting" tools if anyone has any other suggestions!

### DIFF
--- a/include/usgscsm/Distortion.h
+++ b/include/usgscsm/Distortion.h
@@ -1,19 +1,12 @@
 #ifndef Distortion_h
 #define Distortion_h
 
-#include <vector>
 #include <math.h>
-#include <tuple>
 #include <iostream>
+#include <tuple>
+#include <vector>
 
-enum DistortionType {
-  RADIAL,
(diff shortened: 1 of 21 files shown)


kberryUSGS:
Completes part of #304

Issue #305, opened 2020-07-22
Remove computeAllSensorPartials from Frame and Ls sensor model or modify to call parent after logging


kberryUSGS:
It's not needed, as this is implemented in RasterGM. The Ls sensor model's function does add logging, but if we want to keep the logging, we can leave this method in the Ls sensor model, and have its entire contents be (1) logging (2) call the parent's (RasterGM's) version of this function.

jlaura:
Is this closable? Looks like the auto close in #311 just might not have worked?

kberryUSGS:
Yes, thanks! Closed by #311

PR #310, opened 2020-07-24
Updates SAR to use unified ISD take 2

jessemapel:
Fix of #309 by rebasing onto dev.

Adds updated SAR ISD parsing. This still needs to go through a knoten comparison.

PR #311, opened 2020-07-24
More linter-suggested formatting changes, warnings added for non-intersection, and clean up duplicate code

kberryUSGS:
This PR excludes changes to SAR so that they don't conflict with the work that @jessemapel is doing on #310


kberryUSGS:
Closes #305

PR #318, opened 2021-01-07
Added no sharing criteria

jessemapel:
The linescan and framing models errored when you accessed their parameter sharing criteria. This is not what's suppose to happen when parameters are not sharable. Instead, they should return the default SharingCriteria.

This was causing the default GeometricModel::getParameter and GeometricModel::getParameters to raise an exception when they shouldn't.


kberryUSGS (approved):
:+1: This looks good to me!

scsides (approved):
Approved, I'll let you merge when you want to.

PR #326, opened 2021-02-11
Updated state string to begin with the sensor model name as per the CSM standard

### LINKED ISSUE #317: Model states need to start with the model name
According to the CSM API docs, the state string should start the mode name followed by a new line and then the rest of the data. From the API docs section 3.17:

"Sensor model state data is formatted as an ASCII character std::string where the first set of characters in the std::string (up to and including the first newline character) shall be the sensor model name,[...]"

We need to change the state strings so that they are the model name followed by a new line and then the JSON state data.

kberryUSGS:
Fixes #317 by adding the sensor model name to the beginning of the state string for all sensor models. This change is backwards compatible, so existing state strings without the model name still work. Work done with @scsides.


jessemapel (approved):
Looks good. CI is running a bit slow, but this can go in once it passes.

PR #335, opened 2021-04-06
Updated ls to re-interp states

### LINKED ISSUE #334: HiRISE pairs acting very odd (GXP)
@jessemapel Now that I have been able to test these two HiRISE samples in GXP 4.3, I can confirm some strange behavior undoubtable the same as @oleg-alexandrov was seeing.
https://github.com/USGS-Astrogeology/usgscsm/wiki/CSM-Stereo-Testing-Examples

Triolet actually loads into GXP with very reasonable footprints (location is not great but in the right area) and the aspect (width versus width) looks good also. When viewing in stereo, the initial load looks reasonable although larger parallax than normal. But upon moving a little in stereo, one image collapses in Y by about half. So the first image remains the correct aspect and the second smashes in half (in the Y direction) only. Very strange. There is no way to recover from that in trying to force control. 

Jezero also has okay good initial location and footprint overlap looks good. Stereo looks good too (no smashing). Automatic point measure (APM) fails pretty harshly but I can force some manual control and triangulation. Running a DEM now and will check back in the morning. APM "should" work pretty well for this, but maybe the initial locations are just too far off. 

I guess I recommend perhaps testing both Triolet images in a knoten notebook (to see if the returned footprint acts up). I'll be happy to demo in GXP too.

jessemapel:
Fixes #334 

The ideal solution here would be to just adopt the ALE objects as the internal representations, but this is a shorter term patch for the problem. We'll want to re-validate everything in the pipeline before we pull a release off this.


kberryUSGS (approved):
:+1: This looks good and makes sense to me.

PR #340, opened 2021-07-06
Add the ability to apply a transform to a linescan model

oleg-alexandrov:
This makes it possible to apply to a linescan model state a rotation and translation, whose effect is to change the camera position and orientation in ECEF.

This is something I needed in ASP for bundle adjustment. I have a copy of this functionality over there, so I don't need this pull request for my own work, but I thought it would make sense in the linesecan model and could be used by other tools which need to modify this state.

### DIFF
--- a/include/usgscsm/UsgsAstroLsSensorModel.h
+++ b/include/usgscsm/UsgsAstroLsSensorModel.h
@@ -154,6 +154,12 @@ class UsgsAstroLsSensorModel : public csm::RasterGM,
 
   virtual std::string getModelState() const;
 
+  // Apply a rotation and translation to a state string. The effect is
+  // to transform the position and orientation of the camera in ECEF
+  // coordinates.
+  static void applyTransformToState(ale::Rotation const& r, ale::Vec3d const& t,
+                                    std::string& stateString);
+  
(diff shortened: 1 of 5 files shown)


jessemapel (commented):
This looks good but needs tests

include/usgscsm/UsgsAstroLsSensorModel.h
```diff
@@ -971,6 +971,12 @@ class UsgsAstroLsSensorModel : public csm::RasterGM,
   // Compute the determinant of a 3x3 matrix
   double determinant3x3(double mat[9]) const;
 
+  // Apply a rotation and translation to a state string. The effect is
+  // to transform the position and orientation of the camera in ECEF
+  // coordinates.
+ void applyTransformToState(ale::Rotation const& r, ale::Vec3d const& t,
```
jessemapel, line 977: Since this isn't operating on the actual model's state this should be a static method.

oleg-alexandrov:
I made the transform application function static and added a test.

jessemapel (commented):
This looks good to go now. I need to fix our CI as Travis has shut down but once that's done I'll get this merged in.

PR #347, opened 2021-08-09
Added indent and new line to state dumps

### LINKED ISSUE #344: Have the json state be saved on multiple lines
ALE now saves an ISD on multiple lines, but USGSCSM saves the model state as one line. I would think that, as before, it would help with readability if the state was saved on multiple lines as well. Any thoughts? 

BTW, I could not find any function either in USGSCSM or in CSM itself reading a model state from disk. All I can see is a function for reading the ISD from disk, meaning loadImageSupportData(). And apparently a function for saving a state to disk is not present either.

jessemapel:
Fixes #344

### DIFF
--- a/src/UsgsAstroFrameSensorModel.cpp
+++ b/src/UsgsAstroFrameSensorModel.cpp
@@ -752,7 +752,7 @@ std::string UsgsAstroFrameSensorModel::getModelState() const {
        {m_referencePointXyz.x, m_referencePointXyz.y, m_referencePointXyz.z}},
       {"m_currentParameterCovariance", m_currentParameterCovariance}};
 
-  std::string stateString = getModelName() + "\n" + state.dump();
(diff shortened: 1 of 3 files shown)


jessemapel:
As long as the state->model->state tests are working, no additional tests are needed for this as the JSON library handles both adding the indent and parsing it on re-ingest.

jlaura (approved):
From the docs - looks like this is going to indent nicely for objects and then also impact arrays? So will the ephemeris information turn into:

[#,
 #,
 #,
 #]

or just the keys in the object gets pretty printed and the arrays stay is line form?

Either way :+1:.

jessemapel:
Unfortunately, arrays will also get a whole bunch of new-lines. This will make states a good bit larger, but still relatively small.

jlaura (approved):

PR #352, opened 2021-09-16
Add functions for applying transform to SAR and Frame sensor

oleg-alexandrov:
I added functions to apply a transform to model state for SAR and Frame sensors (such logic for Linescan sensors was added a while ago). I put unit tests too. 

There are some other very small fixes, including to dump in some more places the model state on multiple lines. 

I made the function getModelNameFromModelState() in the Sar class static as in the rest of the the classes, which required removing some debug lines, which is not likely to affect anything. 

I don't need these functions for ASP, as for now I use a clone of these, but when a new USGSCSM version is pushed to conda, I will use the functions from here instead of the ones in ASP.

These changes were tested very carefully with real data, though the unit test provided here is a trivial one to just go through the motions.

### DIFF
--- a/include/usgscsm/UsgsAstroFrameSensorModel.h
+++ b/include/usgscsm/UsgsAstroFrameSensorModel.h
@@ -34,8 +34,16 @@ class UsgsAstroFrameSensorModel : public csm::RasterGM,
       double *achievedPrecision = NULL,
       csm::WarningList *warnings = NULL) const;
 
+  static std::string getModelNameFromModelState(const std::string& model_state);
+
   std::string constructStateFromIsd(const std::string &jsonIsd,
                                     csm::WarningList *warnings);
+  
+  // Apply a rotation and translation to a state string. The effect is
+  // to transform the position and orientation of the camera in ECEF
+  // coordinates.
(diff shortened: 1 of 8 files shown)


oleg-alexandrov:
I wonder what is the status on this. These parallel the logic added and accepted into the linescan sensor model for applying a transform to a model state. More details are above.

jessemapel:
Current status on this is that I need to review it but I'm oversubscribed at the moment. I'd like to take some time and look through all the issues you've found with the SAR model under our funded efforts this year as opposed to our general support work.

We haven't started any of that work yet and we have some outstanding projects to tie off so I'm not sure when that will happen. I'd like to get all of this out with the changes in #358

oleg-alexandrov:
Given that the outstanding issues mentioned above have been resolved, can this be accepted at some point? These proposed functions mirror what the linescan sensor already has and do not interfere with other code. 

In a little while we'll move to other projects, and by next year likely things will be harder to merge. 

This logic is meant to apply a transform to a state which should benefit photogrammetry tools. At some point I'll have ASP call them, and I will do another round of testing to ensure all is correct.

BTW, I found that this had a minor conflicts when comparing to the latest code and put a fix.

jessemapel (approved):

Issue #357, opened 2022-01-26
Problems with some LRO NAC datasets


oleg-alexandrov:
I have 75 LRO NAC datasets very close to the South Pole (within 0.2 degrees or so). I made CSM .json files for all of them, and evaluated the accuracy of projecting rays from the ISIS model onto the ground and back to the CSM model and vice versa. For 56 of them the pixel-to-pixel error is under 0.5 pixels. For some, it is 100 pixels, for some it becomes 70,000 pixels.

I must say that I am still using ISIS 5.0.1, ale 0.8.5, and ISISDATA likely a year old, so there is a chance the issue is on my side. I am surprised however, that for most datasets things work out, but for some they don't. 

I am attaching for the record the problematic .cub files, in case somebody has any quick ideas. 

The list below has the  image name and pixel-to-pixel error, and for completeness listed are a handful of datasets for which this is error actually very good. The worst are towards the bottom.

M169576603LE    0.233772
M139831449LE    0.234603
M139831449RE    0.234668
M139729686LE    0.235561
M141059693LE    0.242376
M169488450LE    0.245601
M169474886LE    0.246625
M157830228LE    0.257869
M169468103LE    0.261228
M157830228RE    0.261858
M169502013RE    0.264729
M169556262RE    0.287273
M169549481RE    0.300138
M169549481LE    0.302653
M157809893RE    0.307175
M157843789LE    0.30771
M1315066257LE   0.34057
M169278138LE    0.354418
M169291703LE    0.369411
M139716114RE    0.397297
M169522357LE    0.434125
M169508794LE    0.437499
M150366876RE    3.12818
M148012947LE    3.6963
M110962901RE    7.94198
M113679075LE    15.034
M147992619RE    25.0667
M108946135RE    34.354
M167289893LE    353.674
M121014786LE    747.538
M162587237LE    926.804
M162580453LE    963.458
M121028353LE    975.651
M162553316RE    1233.43
M162499044LE    1593.95
M162492261LE    1858.43
M162485477LE    2138.01
M162478694LE    2404.19
M114859732RE    51393.9
M119582918RE    61134.2
M117241732LE    70079.7

jessemapel:
Is any of this limb data? CSM and ISIS handle the limb quite differently. CSM will give you the closest point on the body to the look vector while ISIS will just return a non-intersection.

I'd also double check your surface model and make sure you're using the same surface model intersection routine.

jessemapel:
[...]

PR #358, opened 2022-01-28
Fix for the linescan sensor, the step size can be too big when iterating

oleg-alexandrov:
This is a fix for the situation when the linescan sensor groundToImage() function fails to converge to desired precision. The problem is that sometimes the step size being used is too big, and it overshoots. The solution is to make the step size smaller when this scenario is detected.

This situation is supposed to happen only for some peculiar sensors. I ran into it with LRO NAC towards the South Pole, as described here: https://github.com/USGS-Astrogeology/usgscsm/issues/357

In situations where things were fine even before, I verified that no changes will happen.

It is not clear how much smaller to try to make the step. Here I chose to make it half the original size when the original one is too big. Presumably, the closer one is to the "optimal" size, the faster the algorithm will converge without overshooting. Yet, it would a fragile thing to do to aim to find the optimal step size, since that is done for just the center pixel when the model is set up. There's always a chance for some other pixel in the sensor this "optimal" step size may be too big.

This fix was validated for 74 LRO NAC images, out of which 19 were failing to converge. Now all converge with pixel differences of under 0.44 pixels when comparing projecting pixels to the ground with the ISIS camera and projecting back with the CSM camera, and vice versa.

### DIFF
--- a/include/usgscsm/UsgsAstroLsSensorModel.h
+++ b/include/usgscsm/UsgsAstroLsSensorModel.h
@@ -125,6 +125,7 @@ class UsgsAstroLsSensorModel : public csm::RasterGM,
   double m_halfTime;
   std::vector<double> m_covariance;
   int m_imageFlipFlag;
+  double m_stepSizeFactor;
 
   std::vector<double> m_sunPosition;
   std::vector<double> m_sunVelocity;
--- a/include/usgscsm/UsgsAstroLsSensorModel.h
+++ b/include/usgscsm/UsgsAstroLsSensorModel.h
@@ -125,7 +125,6 @@ class UsgsAstroLsSensorModel : public csm::RasterGM,
   double m_halfTime;
   std::vector<double> m_covariance;
   int m_imageFlipFlag;
-  double m_stepSizeFactor;
 
   std::vector<double> m_sunPosition;
   std::vector<double> m_sunVelocity;
@@ -977,7 +976,14 @@ class UsgsAstroLsSensorModel : public csm::RasterGM,
 
   // Compute the determinant of a 3x3 matrix
   double determinant3x3(double mat[9]) const;
-
+  
+  // A function whose value will be 0 when the line a given ground
+  // point projects into is found. The obtained line will be
+  // approxPt.line + t.
+  double calcDetectorLineErr(double t, csm::ImageCoord const& approxPt,
+                             const csm::EcefCoord& groundPt,
+                             const std::vector<double>& adj) const;
+  
   csm::NoCorrelationModel _no_corr_model;  // A way to report no correlation
                                            // between images is supported
   std::vector<double>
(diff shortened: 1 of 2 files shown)


oleg-alexandrov:
The previous fix I put in was trying too hard. For the CTX dataset B17_016219_1978_XN_17N282W no amount of halving the step size will convince it to converge to desired precision 0.001. But the first halving made it converge from precision 0.0031396  to 0.00193912.

Hence, a modification is made here, to try to halve the step size just once. That is enough to fix all the LRO NAC issues I found. It will also cause datasets which fail to converge to halve the step size, as CTX above, but again, just once, which hopefully is not a big deal.

A foolproof solution would be in the GroundToImage() function itself for every single pixel. It would adjust dynamically the step size, once the size of m_iTransL is set beforehand. At different pixels different steps may be required, since as the camera flies in space it can change orientation half-way on a whim, so a global solution cannot work.

include/usgscsm/UsgsAstroLsSensorModel.h
```diff
@@ -125,6 +125,7 @@ class UsgsAstroLsSensorModel : public csm::RasterGM,
   double m_halfTime;
   std::vector<double> m_covariance;
   int m_imageFlipFlag;
+  double m_stepSizeFactor;
```
jessemapel, line 128: This needs to be serialized into the model state like the other computed members. It can be an optional value on the read back in to preserve backwards compatibility.

jessemapel (commented):
This change seems reasonable to resolve the issue, but I also think there's a better way to handle this. We're already doing a pseudo-secant method approach where we adjust the step size every iteration. I'd also like to see a unit test case for this situation. I need to look into the LRO NAC data and see what exactly is causing this to overshoot in the iterations.

oleg-alexandrov:
If you would like to test the pseudo-secant method solution you are developing, here are the JSON files for which the current approach fails, before the fix I put in, that is:

[csm_testcases.zip](https://github.com/USGS-Astrogeology/usgscsm/files/7972687/csm_testcases.zip)

I will also suggest examining this CTX dataset: 
[B17_016219_1978_XN_17N282W.8bit.json.txt](https://github.com/USGS-Astrogeology/usgscsm/files/7972721/B17_016219_1978_XN_17N282W.8bit.json.txt)

This problem is a little different. It fails to converge with desired precision 0.001 even with smaller step sizes, which I found rather unexpected.

oleg-alexandrov:
I implemented the secant method for finding the right sensor line in GroundToImage(). It converges for all of a set of 93 LRO NAC polar datasets I tested with. The original method (before this pull request) had trouble with many of them. 

The secant method is also almost 2x faster than the original method, when the original one works. 

This is to be expected from theoretical considerations.  The original approach is a fixed point method. That one normally has linear convergence, and it can diverge even when starting arbitrarily close to fixed point (depending on the derivative at the fixed point). The secant method has superlinear convergence, more exactly, the rate is 1.6. The secant line also directs the updates towards the root even when the two points used to fit it are both on the same side of the root, when the fixed point iteration makes things move further away.

jessemapel (approved):

Issue #350, opened 2021-09-13
imageToRemoteImagingLocus in the SAR sensor


oleg-alexandrov:
The function imageToRemoteImagingLocus() in the SAR sensor does not work correctly. It should return the direction of the vector from getSensorPosition(imagePt) to imageToGround(imagePt), and that holds for the Linescan sensor with 15 digits of precision. Yet, for the SAR sensor these are way different. Here's one example:

Locus function output:                          (0.848551, 0.529113, -0.000527521) 
normalized(ground - sensor_position): (0.970237, -0.201042, -0.134989) 

This was tested with two datasets, LSZ_01636_1CD_XKU_09N120_S1.8bit.cub and LSZ_02331_1CD_XKU_15N120_S1.8bit.cub, and corresponding json files. (Issue https://github.com/USGS-Astrogeology/usgscsm/issues/349 explains how to fetch and prepare the inputs.)

oleg-alexandrov:
Fixed in https://github.com/USGS-Astrogeology/usgscsm/pull/361.

Issue #349, opened 2021-09-13
MiniRF and CSM


oleg-alexandrov:
The USGSCSM SAR plugin has a discrepancy of about 0.5 pixels compared to the ISIS .cub camera model when used with MiniRF. I verified this both with knoten and with ASP's own test tool, for three MiniRF testcases. When running knoten here are the numbers that come out:

isis2csm_data
        diff line  diff sample
count  100.000000   100.000000
mean     0.006334     0.504320
std      0.003996     0.004481
min     -0.000446     0.499011
25%      0.003520     0.500845
50%      0.006340     0.503148
75%      0.009396     0.506966
max      0.013037     0.517591

csm2isis_data
        diff line  diff sample
count  100.000000   100.000000
mean    -0.006299    -0.503728
std      0.003964     0.006325
min     -0.012582    -0.517626
25%     -0.009131    -0.507721
50%     -0.006316    -0.503513
75%     -0.003540    -0.500938
max     -0.000013    -0.490000

Note the second column of numbers, so the "sample" values. 

I used the data from here: https://github.com/USGS-Astrogeology/usgscsm/wiki/CSM-Stereo-Testing-Examples

Particularly, I tried three datasets: LSZ_01636_1CD_XKU_09N120_S1.8bit.cub, LSZ_02331_1CD_XKU_15N120_S1.8bit.cub, LSZ_03204_1CD_XKU_05N120_S1.8bit.cub.

I used these datasets precisely as provided, so without doing spiceinit or re-creating the json files.  (Those operations fail anyway, as described here https://github.com/USGS-Astrogeology/ale/issues/423.)

Note that the corresponding .json files no longer work with the USGSCSM plugin because they are lacking the line: center_ephemeris_time. I added it manually, as the average of first and last ephemeris time. (The precise value does not seem to matter, as altering by a couple of seconds does not change the result.)

As a sanity check, I ran knoten as well on LRO NAC data, for testcase M1167355858LE (which uses the Linescan plugin), and the errors are below 0.025 pixels.

The latest version of knoten source code from github was used with usgscsm 1.5.2. 

I tested only the image to ground and ground to image functions. I vaguely remember I ran into some unrelated issues with the imageToRemoteImagingLocus() function for SAR, which I am not sure that knoten tests.  

Here's the code used to call knoten. Note that the latest "base" and "lro" data were fetched before doing this comparison.

import os, sys

[...]

PR #361, opened 2022-04-12
Bugfix in imageToRemoteImagingLocus

oleg-alexandrov:
This is a fix for the fact that imageToRemoteImagingLocus() was giving an incorrect answer (https://github.com/USGS-Astrogeology/usgscsm/issues/350). 

The problem seems to be that this function goes through some of the same motions as imageToGround() (with different notation), but returns the cross-track vector instead of the vector pointing to the ground.

I could have just pasted the same logic as in imageToGround(), or carefully adjusted the existing  imageToRemoteImagingLocus() code along the same lines, but that would have been just code repetition. So, the change I made is to call the correct function imageToGround() and then finish the rest of the calcuation. I used a height of 0 when calling this, since we don't care to hit a specific height above datum, we just want the direction of the ray towards the ground and a point on it. 

While in theory the imageToGround() function that is now invoked is iterative, it finds the solution in the first iteration, since given how the nadir, across track and along track vectors are defined, if one does carefully the math using those formulas it turns out the exact solution is obtained at once. Running the code verifies that. 

Now imageToGround() is consistent with imageToRemoteImagingLocus() and consistent with the ISIS implementation of imageToGround() and groundToImage(). There's still a small discrepancy somewhere (https://github.com/USGS-Astrogeology/usgscsm/issues/349) but now things are at least in the ballpark.

### DIFF
--- a/src/UsgsAstroSarSensorModel.cpp
+++ b/src/UsgsAstroSarSensorModel.cpp
@@ -543,6 +543,10 @@ csm::EcefCoord UsgsAstroSarSensorModel::imageToGround(
     pointRadius -= (pointHeight - height);
   } while (fabs(pointHeight - height) > desiredPrecision);
 
+  if (achievedPrecision) {
+    *achievedPrecision = fabs(pointHeight - height);
+  }
+
   csm::EcefCoord groundPt(groundVec.x, groundVec.y, groundVec.z);
 
   return groundPt;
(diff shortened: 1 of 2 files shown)


jessemapel:
So this doesn't quite fix the issue with the imaging locus in the SAR. In CSM, the imaging locus is defined as the set of observable points in object space for a given image pixel. So, for a simple pinhole camera it is just a straight line from the sensor focal point to the ground point. For more complex cases like atmospheric defraction or a SAR imager, it's a curve. Here's the relevant text from the CSM API:

> There is a one-to-one mapping from ground space to image space. However, from image to ground space, the mapping is one-to-many. Each pixel in image space corresponds to a set of ground points (for optical imagery, this set of ground points comprises the line of sight). That set of ground points is termed the imaging locus. The imaging locus is useful in a variety of photogrammetric operations. For example, an image-to-ground calculation can be made by finding the intersection of the imaging locus with a ground surface.
> 
> In general, the imaging locus is a complicated path due to the effects of physical phenomena. Therefore, the sensor models return an approximation to the imaging locus termed the proximate imaging locus. The proximate imaging locus consists of a point on the locus together with the direction of the locus at that point. In effect, the proximate imaging locus at a given ground point is the line tangent to the imaging locus at that point. Any arbitrary imaging locus can then be approximated using multiple proximate imaging loci.
> 
> For optical imagery, the imaging locus follows the line of sight from the camera to the ground. Figure depicts the imaging locus for optical imagery. The effects of refraction on the line of sight are greatly exaggerated in this picture. For SAR imagery, the imaging locus follows the curve of intersection of a constant range sphere with the Doppler cone and also includes the effects of refraction.
> 
> In cases where a ground point is not available to compute the proximate imaging locus, a remote imaging locus can be obtained. The remote imaging locus is the line of sight approximated by a straight line. The complicated effects of physical phenomena are ignored. This gives a good approximation of the actual imaging locus and can be used to obtain ground points at which a proximate imaging locus can be computed.

I'm not 100% clear on if the remoteImagingLocus should return just a straight line or if it should  return an arbitrary point on the locus. The second to last paragraph clearly states that a SAR locus is the intersection of the dopplar cone and range sphere (effectively a circle), but the last paragraph says that the remoteImagingLocus function should give a "straight line".

oleg-alexandrov:
Atmospheric refraction is a problem even with optical cameras, so with the linescan and frame sensors. In ASP we have a correction for that, but regretfully you either have to model the atmosphere for each of your bodies very well, and get a curved ray, or one can do some kind of ad-hoc rule, yielding still a straight ray but with a slightly different direction. That can be good enough, since the spacecraft is way up, 600 km, for example, but the ray bends only in the last 20 km or so, at least for Earth.

All we can do is what your text says:  "The complicated effects of physical phenomena are ignored."

The good news is that unless your camera is at a highly oblique angle, any distortion due to the atmosphere (and there is also a velocity aberration distortion because the satellite moves quite fast),  the effect of such things is about the same for all pixels in the camera. So, for the purpose of stereo one can just ignore such high-order effects, do bundle adjustment, which would also correct camera positioning errors, create a DEM, then align that one to your known reference. This will take care of all the minor systematic errors as above.

There is no intrinsic value in precise modeling of the curve emanating from the camera pixel to the ground, as photogrammetry works only with straight rays. 

The resulting expression would be too complicated to be of any use. There is no easy way to model or tabulate Earth's (or Venus's) atmosphere accurately enough for this to work. Earth's atmosphere is also thicker at the equator than the pole, and the way light bends is influenced by temperature, so it may be seasonal, and local fluctuations. 

I think that it is good enough to ensure that the imaging locus is consistent with the image to ground direction.

oleg-alexandrov:
The text you pasted also says: "The imaging locus is useful in a variety of photogrammetric operations. For example, an image-to-ground calculation can be made by finding the intersection of the imaging locus with a ground surface."

So, as long as the image-to-ground function and following the imaging locus ray yield the same results, we are good. Anything else would need curved rays, and those are no-starter in photogrammetry.

jessemapel (approved):
Based on our discussions, for right now let's move forward with assuming it's an optical sensor in the remote imaging locus approximation. Then, we can see if that ends up being a problem when we test on proper data.

PR #363, opened 2022-04-15
Updated for USGS release requirements

jessemapel:
This will require a 1.5.3 release to be valid because of the changes in the license file name and new disclaimer file.


thareUSGS:
updates look good to me.

PR #373, opened 2022-04-20
Add the tool usgscsm_cam_test

oleg-alexandrov:
This is a tool intended to be shipped with the usgscsm library (it will install itself in 'bin'). The goal is for the user to invoke it for basic sanity checks of CSM models. For now, all it can do is load a CSM model, for example, as:

    bin/usgscsm_cam_test tests/data/orbitalSar.json 

when it will produce output:

    Loaded CSM model of type USGS_ASTRO_SAR_SENSOR_MODEL from ../tests/data/orbitalSar.json.

Future proposed functionality:

 - Test projecting rays from the camera to ground and vice-versa, with given sample rate.
 - Load a CSM model state (stored in a .json file, just like an ISD model).
 - Ability to export a CSM model in ISD format to a CSM model state file.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -103,4 +103,10 @@ if(USGSCSM_BUILD_TESTS)
   include(CTest)
   enable_testing()
   add_subdirectory(tests)
+
+  # Test the test_usgscsm_cam_test program
+  add_test(NAME test_usgscsm_cam_test_linescan
+      COMMAND usgscsm_cam_test data/orbitalLineScan.json
+      WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}/tests)
+      
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -80,10 +80,13 @@ target_link_libraries(usgscsm
                       ${ALE_TARGET}
                       nlohmann_json::nlohmann_json)
 
+add_executable(usgscsm_cam_test bin/usgscsm_cam_test.cc)
+target_link_libraries(usgscsm_cam_test
+    usgscsm)
 
 install(TARGETS usgscsm LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR})
 install(DIRECTORY ${USGSCSM_INCLUDE_DIRS} DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
-
+install(TARGETS usgscsm_cam_test RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
 
 # Optional build tests
 option (USGSCSM_BUILD_TESTS "Build tests" ON)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -104,9 +104,4 @@ if(USGSCSM_BUILD_TESTS)
   enable_testing()
   add_subdirectory(tests)
 
-  # Test the test_usgscsm_cam_test program
-  add_test(NAME test_usgscsm_cam_test_linescan
-      COMMAND usgscsm_cam_test data/orbitalLineScan.json
-      WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}/tests)
-      
 endif()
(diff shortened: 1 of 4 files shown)


bin/usgscsm_cam_test.cc
```diff
@@ -0,0 +1,64 @@
...
+  std::string model_file = argv[1];
+  csm::Isd isd(model_file);
+
+  std::vector<std::string> model_types = {"USGS_ASTRO_FRAME_SENSOR_MODEL",
+                                          "USGS_ASTRO_LINE_SCANNER_SENSOR_MODEL",
+                                          "USGS_ASTRO_SAR_SENSOR_MODEL"};
+
+  // Try to load the model
+  bool success = false;
+  UsgsAstroPlugin plugin;
```
jessemapel, line 34: This should go through all possible loaded plugins and not just our plugin.

bin/usgscsm_cam_test.cc
```diff
@@ -0,0 +1,64 @@
...
+
+  if (argc != 2) {
+    std::cerr << "Usage: " << argv[0] << " <model file>" << std::endl;
+    return 1;
+  }
+  
+  std::string model_file = argv[1];
+  csm::Isd isd(model_file);
+
+  std::vector<std::string> model_types = {"USGS_ASTRO_FRAME_SENSOR_MODEL",
```
jessemapel, line 28: This is accessible from the plugin. See Plugin::getNumModels and Plugin::getModelName

oleg-alexandrov:
I tried following this more generic CSM approach yesterday and no plugins were loaded. Now I figured out why. Unless some usgscsm object is actually created in the program, the libusgscsm library fails to get detected as a valid plugin. 

Likely the issue is due to the fact that I link to libusgscsm.so dynamically in CMakeLists.txt, rather than doing some funny business with boost::dll that, if I recall, the CSM spec suggests, but that would just be an extra complication as I see it.

Anyhow, for now I put a fix for that and addressing these issues. The tool prints:

Detected CSM plugin: UsgsAstroPluginCSM
Number of models for this plugin: 3
Loaded CSM model of type USGS_ASTRO_SAR_SENSOR_MODEL from ../tests/data/orbitalSar.json.

jessemapel:
It would be good to also add a test to this. You can have ctest execute commands with

```
add_test(NAME test_usgscsm_cam_test_ls
               COMMAND usgscsm_cam_test data/orbitalLineScan.json
               WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}/tests
```

oleg-alexandrov:
I added a test. I verified that the test passes when it should pass and fails when it should fail.

CMakeLists.txt
```diff
@@ -103,4 +103,10 @@ if(USGSCSM_BUILD_TESTS)
   include(CTest)
   enable_testing()
   add_subdirectory(tests)
+
+  # Test the test_usgscsm_cam_test program
+  add_test(NAME test_usgscsm_cam_test_linescan
+      COMMAND usgscsm_cam_test data/orbitalLineScan.json
+      WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}/tests)
```
jessemapel, line 110: Put this into the CMakeLists.txt file in the tests directory so it's with the other tests.

oleg-alexandrov: done

jessemapel:
This looks like a good starting point, do you want to keep building the functionality on this PR or merge this and add functionality in a separate PR?

oleg-alexandrov:
I think you can pull it in. I also hope you are fine with the plan outlined at the top. Further suggestions on functionality are also welcome. Then I can add the remaining functionality in a separate pull request to make the review simpler, perhaps. 

I also plan to add documentation to the main README about how to use this tool to test a given CSM ISD file and model state file and how to export an ISD to a model state. Would that be a good place for all that?

jessemapel (approved):
Adding all of that in a new PR sounds good to me

PR #377, opened 2022-04-22
Expansion of the usgscsm_cam_test tool and documentation

oleg-alexandrov:


bin/usgscsm_cam_test.cc
```diff
@@ -1,74 +1,211 @@
...
+  std::cout << "Max:    " << errors.back() << "\n";
+  std::cout << "Count:  " << errors.size() << "\n";
+}
+
+double pixDiffNorm(csm::ImageCoord const& a, csm::ImageCoord const& b) {
+  return sqrt((a.line - b.line) * (a.line - b.line) + (a.samp - b.samp) * (a.samp - b.samp));
+}
+
+// Load a CSM camera model from an ISD or state file. Return 0 on success.
+int load_csm_camera_model(std::string const& model_file,
```
jessemapel, line 148: Why return a success/failure code here? If that's the intention, then this should just return a boolean for readability.

bin/usgscsm_cam_test.cc
```diff
@@ -1,74 +1,211 @@
...
+  
+  // It is safe to access non-existent values from a map, the result
+  // will be an empty string
+  opt.model = parsed_options["model"];
+  if (opt.model == "") 
+    print_help_and_exit = true;
 
-  if (argc != 2) {
-    std::cerr << "Usage: " << argv[0] << " <model file>" << std::endl;
+  // Enforce that the sample rate is an integer
```
jessemapel, line 90: I don't think this is required. For example, you could test every half or tenth of a pixel. The for loops will need to be updated to doubles though.

jessemapel (commented):
Looks good, just a few minor comments. It would be good to have a test for the model state dump option. Maybe run it once and dump the state, then run it again and pass in the state file? I'd also like to split these helper functions out into something we can unit test.

bin/usgscsm_cam_test.cc
oleg-alexandrov: This block is no longer in the latest version I put up for review, per:  https://github.com/USGS-Astrogeology/usgscsm/pull/379

jessemapel: I'm confused what the test step size has to do with #379. Can you clarify for me?

oleg-alexandrov: I looked at your red text, but you were talking about the green text.

oleg-alexandrov: It is hard to explain to the user what a sample rate of 10.23 is. Everybody understands what a sample rate of 10 is. If a subpixel offset is desired, that one is done with the --subpixel-offset option. So, if your sample rate is 10, and subpixel offset is 0.23, it will test pixels 10.23, 20.23, etc. What you are proposing amounts to testing pixels 10.23, 20.46, etc. I guess the end results are about the same, but with what I propose things may be easier to explain.

bin/usgscsm_cam_test.cc
oleg-alexandrov: Fixed in the latest commit.

bin/usgscsm_cam_test.cc
jessemapel: Okay that sounds reasonable and the subpixel-offset makes more context this way.

jessemapel (approved):
I think is good to go once the small merge conflict is resolved. I may put in a PR later on to unit test stuff, but I agree. Right now, everything is fully tested by the two tests.

jessemapel:
I'm okay with merging this if the windows json stuff continue to be a problem

oleg-alexandrov:
Thanks. I think you can pull it in. The windows error is because of getopt.h is missing. That's what I use now to parse command line options. I will think of what to do about that. Maybe using boost, but that is not in libusgscsm's dependencies, and I would not want to add it. I'll try to implement a simple command line parser or something.

jessemapel:
We may be able to use something like this depending on licensing https://github.com/Chunde/getopt-for-windows

jessemapel (approved):

PR #378, opened 2022-04-22
Try to make the Windows build work

oleg-alexandrov:
This is an experiment trying to see if the Windows build can be made to work. This should trigger the build.


oleg-alexandrov:
After a renewed attempt I fixed the Windows build. This was actually a real problem on Linux too, JSON was missing its nlohmann_jsonConfig.cmake config file. On my local system it was being pulled from an unrelated location, but that is not behavior we can count on. 

I can't say this is an elegant solution, it looks that one needs to add some supporting macros for JSON, but it seems to work. Any thoughts on pulling this in? 

(The apparently unrelated fix to usgscsm_cam_test.cc included here is because this one stopped compiling all of a sudden. I will make sure this does not cause any merge conflicts in the other pull request where this changed a lot.)

jessemapel (approved):
I think this is a fine workaround. I'll check the feedstock and see if the there's a known issue of the cmake config not being installed correctly.

I made the same updates on my PR to fix the compilation issue, but I can fix the conflicts over there

PR #380, opened 2022-04-22
Remove the dependency on getopt.h

oleg-alexandrov:
I added some simple parsing which assumes each option has only one value. We could incorporate getopt.h into the code but that one is GPL, which can be problematic. This seems to work. Let's see how the windows build is working out.

### DIFF
--- a/bin/usgscsm_cam_test.cc
+++ b/bin/usgscsm_cam_test.cc
@@ -51,7 +51,7 @@ bool parseOptions(int argc, char **argv, Options & opt) {
     printUsage(argv[0]);
     return false;
   }
-  
+
   // Collect the parsed options in a map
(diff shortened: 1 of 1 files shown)


oleg-alexandrov:
I think it made it. Now, it look that the recent logic Jessie put in, related to OrbitalPushFrameSensorModel, is making some tests fail. But nothing related to the tool I am adding.

jessemapel (approved):
I'm not super happy about how fragile the parsing is here. I'm going to put in a new issue to see if there are any header only or lightweight, non-licensing issue, tools we can pull in to make this happen.

oleg-alexandrov:
Thanks for the merge. Discussion to be continued at https://github.com/USGS-Astrogeology/usgscsm/issues/381.

PR #382, opened 2022-04-26
SAR model: Fix an infinite loop issue

oleg-alexandrov:
I found a situation where the imageToGround() function in the SAR model got in an infinite loop because it could not bring the desired precision to under 1e-10 and 1e-11 was asked. The solution is to limit the number of iterations instead of using a "while" loop.

As an aside, normally we'd expect machine precision in such calculation, but given that it has to deal with the body radius in meters, which is 1737400.0, likely some precision loss happens. 

As a second aside, it looks to me that iterations are not even necessary here. This calculation seems to intersect a ray with an ellipsoid, and that one has an exact solution. The loop here hits an error of 1e-10 in the first iteration and then just goes around that not making any further progress. But this would require some more careful math than what I did to prove for sure, as I did not follow precisely the slant range calculation and the logic of decomposing a vector using a basis of what this code calls uHat, vHat, and tHat.

### DIFF
--- a/src/UsgsAstroSarSensorModel.cpp
+++ b/src/UsgsAstroSarSensorModel.cpp
@@ -524,12 +524,15 @@ csm::EcefCoord UsgsAstroSarSensorModel::imageToGround(
   //   The cross-track unit vector is orthogonal to the position so we ignore it
   double nadirComp = dot(spacecraftPosition, tHat);
 
-  // Iterate to find proper radius value
+  // Iterate to find proper radius value. Limit the number of iterations
+  // because the desired precision may not drop beyond 1e-10 no matter
+  // how many iterations one does. As it was observed, this in fact
+  // converges at the first iteration.
   double pointRadius = m_majorAxis + height;
   double radiusSqr;
   double pointHeight;
   csm::EcefVector groundVec;
-  do {
+  for (int it = 0; it < 10; it++) {
     radiusSqr = pointRadius * pointRadius;
     double alpha =
         (radiusSqr - slantRange * slantRange - positionMag * positionMag) /
@@ -542,7 +545,10 @@ csm::EcefCoord UsgsAstroSarSensorModel::imageToGround(
     pointHeight = computeEllipsoidElevation(
         groundVec.x, groundVec.y, groundVec.z, m_majorAxis, m_minorAxis);
     pointRadius -= (pointHeight - height);
-  } while (fabs(pointHeight - height) > desiredPrecision);
+
+    if (fabs(pointHeight - height) <= desiredPrecision)
+      break;
+  }
 
   if (achievedPrecision) {
     *achievedPrecision = fabs(pointHeight - height);


jessemapel (commented):
Under what conditions did you find this converging in a single iteration? This should converge in a single iteration of the target is a sphere, but not if it's an ellipsoid. Here's the snippet of math from Randy that this is based on:

```
The ground point X is defined implicitly by these 3 simultaneous
equations:
-2.0 * (Xsc(t_min) - X) . V(t) / ( |Xsc(t_min) - X| * WAVELENGTH )= df = 0
|Xsc(t_min) - X| = r_slant
X . X = R^2
where R is the absolute radius at the point of projection.  If we are
projecting the image onto a spherical reference surface, we plug in
a fixed value for R and we are done when we solve these 3 equations.
These equations are quadratic in the coordinates of X.  They are
similar to the equations that arise for projecting an optical
pushbroom scanner, but not (except for the 3rd) identical. Fortunately
they are simple enough we can solve them explicitly by using the
quadratic formula.  The ± sign in the formula corresponds directly
to the look direction left or right, as we shall see.

I previously worked out the general case (needed for Magellan and
Cassini) of solving these equations for any value of df.  The 
solution for df = 0 is simpler and is given here.  The first of
the three equations can be simplified by ignoring the multiplicative
factors (they don't change where the zero value is).  It's also
useful to use the unit vector in the direction of the velocity,
Vhat = V(t)/|V(t)|.  The three equations simplify to: 
|X|^2 = R^2
|X - Xsc|^2 = r_slant^2
(X - Xsc) . Vhat = 0
Now, it is convenient to start by constructing the natural basis 
system in which to find X-Xsc.  The third unit vector in this 
system is Vhat, the first (That) is roughly along C but orthogonal to 
Vhat, and the second (Uhat) is orthogonal to the first two.
Starting from here, this is how we actually calculate, so I will
indent the equations to make them stand out.

	Vhat = V(t) / |V(t)|
	That = (Xsc - (Xsc . Vhat) Vhat) / |Xsc - (Xsc . Vhat) Vhat|
	Uhat = Vhat x That

where x is the vector cross product.  Then, since the surface point
lies in the plane of the circle, we can define the solution as

X - Xsc = alpha That + beta Uhat

with no Vhat component, for some scalars alpha and beta
(not yet known, but they soon will be).  Substituting this 
into the first two equations for X gives

alpha^2 + beta^2 = sigma^2
|alpha That + beta Uhat + Xsc|^2 = R^2

but Xsc = ct That + cv Vhat where 

	ct = Xsc . That
	cv = Xsc . Vhat
	c^2 = ct^2 + cv^2 = |Xsc|^2

because Xsc has no component in the Uhat (along track) 
direction.  Thus, we have

(alpha + ct)^2 + beta^2 + cv^2 = R^2.

Subtracting the other equation from this and rearranging,
we get the explicit solution for alpha:

	alpha = (R^2 - r_slant^2 - c^2) / 2 ct

Finally, the remaining equation gives

	beta = ± sqrt(r_slant^2 - alpha^2)

If the argument of the square root is negative, there is no
real solution and the range circle does not intersect the ground,
so there is no ground point.  In this case we have to exit with
an error condition.  If there is a solution, the + sign
corresponds to right-looking and the - sign to left-looking
radar direction.  Then, finally, the ground point is

	X = alpha That + beta Uhat + Xsc

If the goal is to project the image onto an ellipsoid or a DTM,
we must iterate to find a point where the solution for step 3 is
consistent with the local radius R at that point.  The following 
very simple approach is more or less the state of the art in ISIS.
It works if the surface slopes are small, so that there is no layover
at the scale resolved by the DTM and at the incidence angles that
are relevant.  We proceed through step 3 with a guess for R.
Then calculate the latitude and longitude of the solved-for point X
and evaluate the ellipsoid or DTM radius at that lat and lon.
Iterate evaluating X(R) and R(X) until convergence reaches some
tolerance.
```

oleg-alexandrov:
> Under what conditions did you find this converging in a single iteration?

For the Moon. I am fine with keeping the iterations, as long as there's a limit to their number.

PR #385, opened 2022-05-02
Speedup in Linescan model by avoiding unneeded computations

oleg-alexandrov:
I made a 23% speed improvement in the linescan sensor logic, without changing the results. This relies on avoiding computing the interpolated velocity and applying adjustments when those are not used by the caller.

To verify this, run:

cd usgscsm/build
time ./usgscsm_cam_test --model ../tests/data/toughLroNacLineScan.json --sample-rate 1

before and after applying the fix. On my machine the run time is 155 seconds before the fix, and 118 seconds after it. Granted, not a huge improvement, but it is notable, and the code changes are small.

### DIFF
--- a/include/usgscsm/UsgsAstroLsSensorModel.h
+++ b/include/usgscsm/UsgsAstroLsSensorModel.h
@@ -953,7 +953,9 @@ class UsgsAstroLsSensorModel : public csm::RasterGM,
                              const double& desired_precision,
                              csm::WarningList* warnings = NULL) const;
 
-  // determines the sensor velocity accounting for parameter adjustments.
+  // Determines the sensor position and velocity accounting for
+  // parameter adjustments. Use calc_vel = false if the velocity is
+  // not needed.
   void getAdjSensorPosVel(const double& time, const std::vector<double>& adj,
                           double& xc, double& yc, double& zc, double& vx,
                           double& vy, double& vz, bool calc_vel = true) const;
--- a/include/usgscsm/UsgsAstroLsSensorModel.h
+++ b/include/usgscsm/UsgsAstroLsSensorModel.h
@@ -956,7 +956,7 @@ class UsgsAstroLsSensorModel : public csm::RasterGM,
   // determines the sensor velocity accounting for parameter adjustments.
   void getAdjSensorPosVel(const double& time, const std::vector<double>& adj,
                           double& xc, double& yc, double& zc, double& vx,
-                          double& vy, double& vz) const;
+                          double& vy, double& vz, bool calc_vel = true) const;
 
   // Computes the imaging locus that would view a ground point at a specific
   // time. Computationally, this is the opposite of losToEcf.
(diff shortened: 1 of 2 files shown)


include/usgscsm/UsgsAstroLsSensorModel.h
```diff
@@ -956,7 +956,7 @@ class UsgsAstroLsSensorModel : public csm::RasterGM,
   // determines the sensor velocity accounting for parameter adjustments.
   void getAdjSensorPosVel(const double& time, const std::vector<double>& adj,
                           double& xc, double& yc, double& zc, double& vx,
-                          double& vy, double& vz) const;
+                          double& vy, double& vz, bool calc_vel = true) const;
```
jessemapel, line 959: Update the function doc string to explain the new calc_vel parameter, mostly why you would use it.

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -1938,6 +1941,25 @@ void UsgsAstroLsSensorModel::getAdjSensorPosVel(const double& time,
...
                  time, 3, nOrder, sensPosNom);
+
+  // Avoid computing the velocity and adjustments, if not needed, as those
+  // take up at least half of this function's time.
+  bool has_adj = false;
+  for (size_t it = 0; it < adj.size(); it++) {
+    if (adj[it] != 0) {
+      has_adj = true;
+    }
+  }
```
jessemapel, line 1952: This can be replaced with `bool has_dj = std::all_of(adj.begin(), adj.end(), [](double val){return val == 0.0});` I imagine the compiler will optimize both of these implementations away though.

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -1938,6 +1941,25 @@ void UsgsAstroLsSensorModel::getAdjSensorPosVel(const double& time,
...
+  if (!has_adj && !calc_vel) {
+    xc = sensPosNom[0];
+    yc = sensPosNom[1];
+    zc = sensPosNom[2];
+    vx = 0.0;
+    vy = 0.0;
+    vz = 0.0;
+    return;
```
jessemapel, line 1960: Add a logging message here before the return

jessemapel (commented):
I'm curious how much time is taken by the lagrangeitnerp function vs applying the adjustments.

src/UsgsAstroLsSensorModel.cpp
oleg-alexandrov: I think that lambda functions are kind of terrible for readability. They have their uses, such as being able to quickly cook up a function object to pass to another function, but in this case the simple loop checking for 0 seems to be much easier to understand than this. And yeah, under the hood likely the performance is the same.

oleg-alexandrov:
> I'm curious how much time is taken by the lagrangeitnerp function vs applying the adjustments.

It is the lagrange interpolation which takes up almost all the time. I will look later into optimizing that too, as there are too many repeated multiplications which could be done just once.

If the adjustments are non-zero, one has to compute the velocity in order to find the adjusted position. So this optimization trick works only if both there is no need for velocity and there are no adjustments. Hence, these can't be separated.

src/UsgsAstroLsSensorModel.cpp
oleg-alexandrov: Added.

oleg-alexandrov:
I addressed the above.

jessemapel (approved):

PR #387, opened 2022-05-03
Add an initial projective approximation for linescan camera

oleg-alexandrov:
I implemented an initial guess ground-to-image function for the linecan model which is a projective transform, instead of the existing linear transform. The hope is that this makes more sense geometrically when it comes to projections into camera, so it would fit better the function we want. It is also not that nonlinear to be concerned about it doing funny things when out of range.

With the difficult LRO NAC dataset from the repo, the guessed line using this initial guess projective transform differs from the
eventual computed value of this line by 0.69 pixels.  Before, with the linear approximation, the discrepancy was 82.7 pixels.

With a CTX dataset, the mean line error was 160.4 pixels before, and is 0.85 pixels now.

This does not in a dramatic speed improvement though, because the secant method converges so fast even with a bad initial guess. For LRO NAC (when the sample rate in usgscsm_cam_test is 1) the runtime goes from 122 to 108 seconds, which is a 11% speedup, while for CTX it goes from 82 to 70 seconds, which is a 14% speedup (here I used sample rate of 2).

Anyhow, this is a noticeable improvement, verified with two different datasets, so it is not too bad. Likely there's only very limited room for further improvement by fitting a fancier transform.

A couple more notes. I made USGSCSM depend on Eigen, which is not a big deal, as ALE already depends on Eigen and USGSCSM depends on ALE. I tested that the code compiles fine with both internal and external dependencies.

Also, Eigen is a hog and can slow down compilation speed (from 30 to 45 seconds on my machine for the linescan class). For that reason I isolated the Eigen logic to a single .cpp file to avoid this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,6 +32,9 @@ if(USGSCSM_EXTERNAL_DEPS)
   # Nlohmann JSON
   find_package(nlohmann_json REQUIRED)
 
+  # Eigen
+  find_package(Eigen3 3.3 REQUIRED NO_MODULE)
+   
   # ALE
   find_package(ale REQUIRED)
   set(ALE_TARGET ale::ale)
@@ -53,8 +56,11 @@ else()
   set(ALE_BUILD_TESTS OFF)
   add_subdirectory(ale)
   set(ALE_TARGET ale)
-endif(USGSCSM_EXTERNAL_DEPS)
 
+  # Use Eigen included with ALE
+  add_library (Eigen3::Eigen ALIAS eigen)
+  set(EIGEN3_INCLUDE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/ale/eigen)
+endif(USGSCSM_EXTERNAL_DEPS)
 
 add_library(usgscsm SHARED
             src/UsgsAstroPlugin.cpp
@@ -63,7 +69,8 @@ add_library(usgscsm SHARED
             src/UsgsAstroLsSensorModel.cpp
             src/UsgsAstroSarSensorModel.cpp
             src/Distortion.cpp
-            src/Utilities.cpp)
+            src/Utilities.cpp
+            src/EigenUtilities.cpp)
 
 set_target_properties(usgscsm PROPERTIES
     VERSION ${PROJECT_VERSION}
@@ -71,7 +78,8 @@ set_target_properties(usgscsm PROPERTIES
 )
 
 set(USGSCSM_INCLUDE_DIRS "${CMAKE_CURRENT_SOURCE_DIR}/include/usgscsm"
-                         "${CMAKE_CURRENT_SOURCE_DIR}/include/")
+                         "${CMAKE_CURRENT_SOURCE_DIR}/include"
+                         "${EIGEN3_INCLUDE_DIR}")
 
 target_include_directories(usgscsm
                            PUBLIC
--- a/src/EigenUtilities.cpp
+++ b/src/EigenUtilities.cpp
@@ -56,18 +56,6 @@ void usgscsm::computeBestFitProjectiveTransform(std::vector<csm::ImageCoord> con
   // https://eigen.tuxfamily.org/dox/group__LeastSquares.html
   Eigen::VectorXd u = M.colPivHouseholderQr().solve(b);
 
-#if 0
-  // Verification
-  for (int it = 0; it < numPts; it++) {
-
-    double x = groundPts[it].x, y = groundPts[it].y, z = groundPts[it].z;
-    double r = imagePts[it].line, c = imagePts[it].samp;
- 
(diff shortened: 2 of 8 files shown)


CMakeLists.txt
```diff
@@ -63,15 +69,17 @@ add_library(usgscsm SHARED
...
 
 set_target_properties(usgscsm PROPERTIES
     VERSION ${PROJECT_VERSION}
     SOVERSION 1
 )
 
 set(USGSCSM_INCLUDE_DIRS "${CMAKE_CURRENT_SOURCE_DIR}/include/usgscsm"
-                         "${CMAKE_CURRENT_SOURCE_DIR}/include/")
+                         "${CMAKE_CURRENT_SOURCE_DIR}/include"
+                         "${EIGEN3_INCLUDE_DIR}")
```
jessemapel, line 82: Link Eigen as a target instead of setting variables like this:

https://eigen.tuxfamily.org/dox/TopicCMakeGuide.html

src/EigenUtilities.cpp
```diff
@@ -0,0 +1,75 @@
...
+  // Verification
+  for (int it = 0; it < numPts; it++) {
+
+    double x = groundPts[it].x, y = groundPts[it].y, z = groundPts[it].z;
+    double r = imagePts[it].line, c = imagePts[it].samp;
+ 
+    double r2 = (u[0] + u[1] * x + u[2] * y + u[3]  * z) / (1 + u[4]  * x + u[5]  * y + u[6]  * z);
+    double c2 = (u[7] + u[8] * x + u[9] * y + u[10] * z) / (1 + u[11] * x + u[12] * y + u[13] * z);
+  }
+#endif
```
jessemapel, line 66: Remove this section. If we want to have some sort of debug output here, use a logger

jessemapel (commented):
There needs to be a test for the projective transform computation function

CMakeLists.txt
jessemapel: Eigen also needs to be added to the environment file

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -2136,15 +2139,121 @@ void UsgsAstroLsSensorModel::computeLinearApproximation(
...
+  usgscsm::computeBestFitProjectiveTransform(ip, gp, m_projTransCoeffs);
+  m_useApproxInitTrans = true;
+  
+  MESSAGE_LOG("Completed createProjectiveApproximation");
+}
+
+//***************************************************************************
+// UsgsAstroLineScannerSensorModel::createLinearApproximation
+//***************************************************************************
+void UsgsAstroLsSensorModel::createLinearApproximation() {
```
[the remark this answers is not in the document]
oleg-alexandrov: This function looks like solid code which performs very well. It could be of some use perhaps, one day, or if the proposed projective transform has issue for some quirky sensor (I don't anticipate that, but who knows).  I could factor this out to Utilities.cpp, together with the formula for the determinant, those are too low-level to keep in this sensor's implementation.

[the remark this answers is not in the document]
oleg-alexandrov: In retrospect, I can wipe this old code. It looks like too much work to refactor it. If we need a best-fit linear transform going forward, the code for the projective transform can be tweaked a little to exclude the denominators, and then that will give a linear transform.

CMakeLists.txt
oleg-alexandrov: Added.

oleg-alexandrov: Linking Eigen is not the issue. It is looking up the header which is a problem. I tried yesterday along the lines of what you suggest and it was not picking up the include directory. The way I put it above works both with internal and external dependencies.

src/EigenUtilities.cpp
oleg-alexandrov: I moved this to the unit test.

oleg-alexandrov:
I added a unit test, removed the debug logic, wiped the old transform, and all the builds and tests pass.

jessemapel (approved):
I'm not jazzed about not linking Eigen as a target but I can come back later with a PR to change that

oleg-alexandrov:
I think linking to Eigen is not needed because a header file has the implementation. I used the same logic as what is in ALE.

oleg-alexandrov:
The "target_link_libraries" logic should not help. Its doc says: "Specify libraries or flags to use when linking a given target and/or its dependents. ". This should not affect the problem I was seeing with finding headers which was happening before linking. 

What you will need, I think is "target_include_directories". Normally even this should not be needed, as "find_package" should already set the header, but that was not happening.

Anyhow, I spent a couple of hours yesterday and today to just get things to build. Now things work. If you are feeling adventurous and want to revisit this, one needs to ensure that the build works with both internal and external Eigen (since different logic is used for each), and then see if anything breaks after pushing things.

Issue #386, opened 2022-05-02
More documentation needed


oleg-alexandrov:
This repo needs some more documentation. Some things I can think about are:

 - How an ISD (camera model file) is created, and mention the occasional gotcha when it comes to loading kernels from .cub files rather than quiering the spice. 
 - Mention the WAC sensor (when ready)
 - WAC workflow (this is trickier because the image needs preprocessing)
 - Ideally a little writeup for each sensor, with a sample dataset, perhaps
 - Not sure what to say about the CSM philosophy, so using boost::dll to load the library.

jessemapel:
A whole bunch of stereo testing examples can be found here:

https://github.com/USGS-Astrogeology/usgscsm/wiki/CSM-Stereo-Testing-Examples

There's also test results on Knoten

https://github.com/USGS-Astrogeology/knoten

Some of the notebooks also have processing steps for going from PDS data to sensor model

jessemapel:
The ALE README has some basic instructions for load/loads

https://github.com/USGS-Astrogeology/ale

The ALE docs also have some more information about data setup

https://ale.readthedocs.io/en/latest/tutorials/quick_start.html

jessemapel:
ASP's CSM documentation

https://stereopipeline.readthedocs.io/en/latest/examples.html#community-sensor-model-csm

oleg-alexandrov:
The above doc has CSM examples for linescan and SAR. I just added another one for the Frame sensor, at: https://stereopipeline.readthedocs.io/en/latest/examples.html#example-using-the-usgs-csm-frame-sensor.

jessemapel:
LRO WAC instructions https://github.com/USGS-Astrogeology/usgscsm/issues/365#issuecomment-1115097769

PR #396, opened 2022-05-13
Initial user and instrument docs.

AustinSanders:
Partially addresses #386

As it stands, this documentation includes:
1. instructions for generating and validating an ISD, including how to handle situations in which ale doesn't find a kernel set, i.e. getting kernels from a spiceinit'd cub
2. processing instructions for WAC
3. a (very) brief summary of each sensor for which ALE drivers exist.  These could use additional information, but I wasn't sure what information was desirable in this context
4. a few processing examples.  I'd like to get an example for each supported driver.  I'm running through knoten + other examples, but I'm going through the process of making sure they actually work + trying to sniff out code rot.

### DIFF
--- a/docs/source/sensors/Cassini.rst
+++ b/docs/source/sensors/Cassini.rst
@@ -1,0 +1,39 @@
+Cassini ISS
+===========
+The Cassini Imaging Subsystem comprises two instruments -- a narrow angle camera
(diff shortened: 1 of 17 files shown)


jessemapel (approved):
I think this is ready to get merged. It's not a complete set of documentation, but it's a whole lot and a very good base to build off of.

PR #405, opened 2022-06-15
GXP State File Read (*.sup) Integration

### LINKED ISSUE #400: GXP state file read (*.sup) added to test utility
background:
GXP saves camera state as a single string near the bottom of its' "sup" (supplement) file per image. The .sup is only created when a triangulation is initialized and again after each triangulation solve (bundle adjustment). Three new files have been uploaded to a gxp_testing branch to facilitate this issue (files ESP_037330*.*) : https://github.com/USGS-Astrogeology/usgscsm/tree/gxp_testing/tests/data 

1. Original ALE Image Support Data (ISD) in JSON format, [ESP_037330_1990_REDmos_hijitreged_ale.json](https://github.com/USGS-Astrogeology/usgscsm/blob/gxp_testing/tests/data/ESP_037330_1990_REDmos_hijitreged_ale.json)
2. GXP output for triangulation initialization (in GXP .sup format). [ESP_037330_1990_REDmos_hijitreged.preTriang.sup](https://github.com/USGS-Astrogeology/usgscsm/blob/gxp_testing/tests/data/ESP_037330_1990_REDmos_hijitreged.preTriang.sup)
3. GXP output from post triangulation adjustment (in GXP .sup format). [ESP_037330_1990_REDmos_hijitreged_postTriang.sup](https://github.com/USGS-Astrogeology/usgscsm/blob/gxp_testing/tests/data/ESP_037330_1990_REDmos_hijitreged_postTriang.sup)

From the .sup files, we are only interested in the SENSOR_STATE USGS_ASTRO_LINE_SCANNER_SENSOR_MODEL line. This again contains one giant camera state string. It seems to have an odd or special character "" which may need to stripped before attempting to load into usgscsm.

---

As a user I would like the usgscsm library to be able to parse a GXP .sup file to load the camera state.  The goal is to allow sharing between GXP and other CSM applications like ASP and ISIS. The usgscsm camera state would need to open GXP .sup file, recognize it is a GXP .sup file, find/parse the camera state line, and load it into usgscsm.

definition of done (being able to run):
`usgscsm_cam_test --model ESP_037330_1990_REDmos_hijitreged_postTriang.sup`

If loading is possible, then I would also like to see testing from within usgscsm test application (usgscsm_cam_test.cc). Examples for using the testing application are already available in the repo.

If it is not possible to load a camera state from a .sup file, please document the issues here. For example, is the camera state mangled or other issue.

This is currently a one-way street. We are not attempting to support writing camera state to a GXP .sup file (just reading).

Resources:
1. original test [stereo-pair](https://pdsimage2.wr.usgs.gov/pub/pigpen/CSM_stereo_testing_sets/HiRISE/Mars2020_Jezero_N_IAU_cub.zip), from [wiki](https://github.com/USGS-Astrogeology/usgscsm/wiki/CSM-Stereo-Testing-Examples).
2. Introduction to [usgscsm camera state](https://github.com/USGS-Astrogeology/usgscsm#camera-model-format-and-model-state)
3. Introduction to [ASP's camera state](https://stereopipeline.readthedocs.io/en/latest/examples.html#exporting-csm-model-state-after-bundle-adjustment-and-alignment) capabilities 
4. testing [docs](https://github.com/USGS-Astrogeology/usgscsm/blob/dev/docs/source/tools/usgscsm_cam_test.rst)

amystamile-usgs:
Adds the ability to input a GXP .sup file as a camera state to be used in usgscsm_cam_test. 

Unit tests added to test added utility functions and added plugin test to test reading in sup files as camera state.

Closes #400


tests/data/hello.json
```diff
@@ -0,0 +1 @@
+Hello
```
jlaura, line 1: This is not valid JSON. Should it be?

amystamile-usgs: This is file is just used to test readFileInString so it doesn't necessarily have to be in json format. But I can see how this could be confusing. I think adding in some sort of mock json format would be good here.

jlaura: Literally just wrapping the string in double quotes makes this valid JSON. 

"Hello" is valid. Would that work?

PR #415, opened 2022-08-08
Test reverting #385

jessemapel:
We're seeing triangulation errors between some versions of the code and this is one of the significant changes so testing a revert on this to see if it helps.


jessemapel:
This does appear to be causing the issue, I'm going to dig in a little farther and see if there is anywhere that it's not computing velocities that it should be.

PR #416, opened 2022-08-08
Test Revert #387

jessemapel:
We're seeing triangulation errors between some versions of the code and this is one of the significant changes so testing a revert on this to see if it helps.


jessemapel:
Thankfully this isn't causing the problem.

PR #417, opened 2022-08-08
Checked for set adjustments in getAdjSensorPosVel when determining if velocity is needed

jessemapel:
The change in #385 only checks for adjustments being passed into the function, it doesn't check for adjustments that already exist in the model.

This resulted in the velocity not being computed when adjustments were set in the model and then it cannot compute the in-track.cross-track coordinate system properly.

### DIFF
--- a/src/UsgsAstroLsSensorModel.cpp
+++ b/src/UsgsAstroLsSensorModel.cpp
@@ -1946,7 +1946,7 @@ void UsgsAstroLsSensorModel::getAdjSensorPosVel(const double& time,
   if (!calc_vel) {
     bool has_adj = false;
     for (size_t it = 0; it < adj.size(); it++) {
-      if (adj[it] != 0) has_adj = true;
+      if (getValue(it, adj) != 0) has_adj = true;
     }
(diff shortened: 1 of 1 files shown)


thareUSGS:
The code update looks good and appears to fix the issue. I can confirm using the artifact v0.1.0579 in GXP, both triangulation and stereo-correlation look good with this update.  Feel free to merge.

oleg-alexandrov:
Oh boy. So there's the model, there are preexisting adjustments to it, which are normally 0, and there's yet another set of variable adjustments one may apply on top of those adjustments.  

The names m_currentParameterValue and the function getValue() aren't too helpful in understanding what is going on either. Presumably these things are backdoors for GXP and bundle adjustment. 

Good catch.

jessemapel:
The m_currentParameterValue stuff is for bundle adjustment. The generic adjustments vector is for computing sensor partials using difference approximations; it allows you to tweak the parmeter values without changing the model state.

jessemapel:
I'm also not thrilled with all of the names. This is hold over from BAE's code.

PR #419, opened 2022-08-18
Apply transform to reference point in PushFrameSensor and spellcheck

oleg-alexandrov:
Somehow the PushFrame sensor model is missing a minor bit of logic present for the other models (Frame, Linescan SAR). When a transform is applied to the cameras, the ground reference point should move accordingly, as the camera centers (and orientations and velocities) move. 

Likely this does not matter much, if at all, as the reference point is only used to set up m_gsd and other such things when the model state is created, and it is not used later during actual image-to-ground and ground-to-image operations, but likely it is best to keep it consistent with the rest of the model.


jessemapel (approved):
I'm not sure how much this impacts, but it's good for consistency. I still don't 100% understand what the reference point gets used for by consuming apps. We choose it rather arbitrarily.

Issue #376, opened 2022-04-22
Broken link


oleg-alexandrov:
The link https://github.com/USGS-Astrogeology/pfeffernusse in the READMe file is broken.

jessemapel:
hmm we had to pull that private for compliance reasons, I'm not sure what to do here besides just pull the link for now. @acpaquette @Kelvinrr thoughts?

oleg-alexandrov:
Fixed in https://github.com/USGS-Astrogeology/usgscsm/pull/427.

PR #427, opened 2022-10-20
Removed link to pfefferneuse

jessemapel:
That repo is now internal and likely will be replaced with something else once we have the SpiceQL service sorted out.


jessemapel:
Fixes #376

oleg-alexandrov:
Never have I seen such massive failure only as the result of some tiny doc text being added!

jessemapel:
Something seems to have broken with our compiler compatibility. I'm looking into what happened.

oleg-alexandrov:
I guess this can be pulled in as well. That dead link has been there for a while.

jessemapel:
I'll get this rebased and re-check it before merging.

PR #428, opened 2022-10-25
Initialize m_startingDetectorLine

oleg-alexandrov:
This is a housekeeping fix. The variable m_startingDetectorLine was not being initialized during the reset() call. Normally that's not an issue if the linescan model is being read from an isd, but may be an issue if it is created from scratch. Now this is also going to be consistent with  m_startingDetectorSample (line right above).


jessemapel:
@oleg-alexandrov Can you pull in the most recent changes so CI runs on this. It should be totally fine, but I just want to double check.

oleg-alexandrov:
I did.

jessemapel (approved):

PR #429, opened 2022-10-26
Fixing json version ABI

jessemapel:
The nlohmann JSON 3.11 ABI is not compatible with 3.10. So, our external ALE build that was built againt 3.10 is causing linking errors.


jessemapel:
Also fixing this on the ALE feedstock with https://github.com/conda-forge/ale-feedstock/pull/42

acpaquette (approved):

PR #436, opened 2023-03-01
Line Scan groundToImage Fix

acpaquette:
When converging on an approximate line, when calculating for the error in the detector distortion was not being applied.

On cameras with fairly minor distortion this results in fairly minimal difference in resulting line. On camers, like kaguyatc, with more prevalent distortion, this caused the line to be off by as much as 2 pixels.

Related to this ALE issue: https://github.com/USGS-Astrogeology/ale/issues/508

### DIFF
--- a/src/UsgsAstroFrameSensorModel.cpp
+++ b/src/UsgsAstroFrameSensorModel.cpp
@@ -153,6 +153,11 @@ csm::ImageCoord UsgsAstroFrameSensorModel::groundToImage(
   // Camera rotation matrix
   double m[3][3];
   calcRotationMatrix(m, adjustments);
+  MESSAGE_LOG(
+      spdlog::level::trace,
+      "Calculated rotation matrix [{}, {}, {}], [{}, {}, {}], [{}, {}, {}]",
+      m[0][0], m[0][1], m[0][2], m[1][0], m[1][1], m[1][2], m[2][0], m[2][1],
+      m[2][2]);
 
   // Sensor position
   double undistortedx, undistortedy, denom;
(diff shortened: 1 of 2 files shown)


acpaquette:
@oleg-alexandrov Would you mind checking this out? I want to make sure that this seems reasonable with the results I am seeing. I also tested with MROCTX and got just as good if not better line estimates.

It is likely that this will make the estimation process slightly slower but negligibly so(?)

oleg-alexandrov (approved):
The key change appears on line 2807, with the rest being new log messages. Looks like the right thing to me, though of course careful testing is needed. I guess detectorView[0] is undistorted X, and now it is replaced with distortedFocalLengthX.  The function applyDistortion() should normally be cheap, just some polynomial calculation, unless one deals with distortion models where undistortion operation is a polynomial while distortion operation is an interactive method.

acpaquette:
Right, it looks like both KaguyaLISM and LroNAC use an iterative solve method so some time cost will be incurred on both of those instruments

oleg-alexandrov:
I guess one can time things to see if the slowdown is large. If it is, longer term, given that these sensors have rather weak distortion, it is possible to fit to this distortion model another one which operates in the reverse direction. Not something one would view as a high priority.

acpaquette:
@oleg-alexandrov Doing some experiments in python comparing times in LRONAC, MRO CTX and KaguyaTC. This will add between sub 1 and ~15 seconds to a full image back projection in USGSCSM. Which for something like LRONAC having 51200 lines and 5064 samples is fairly negligible.

In terms of accuracy gain/loss. LRONAC loses no accuracy on the back projected line and MRO CTX loses an average of 0.0002 pixel accuracy on the back projected line.

Overall this change seems reasonable, having minor decreases to performance with negligible accuracy lose in the more tested line scan instruments. All to gain +2 pixel accuarcy in KaguyaTC

oleg-alexandrov:
This is good to know, thank you. So, this fixes the issue where a prime suspect was the projective vs linear initial guess transform? Did you have to switch back there to the linear initial guess? 

This is an important fix. The error due to incorrectly modeled distortion can be worse than 2 pixels, if the line sensor is is not precisely where the optical center is, and many cameras do have more than one sensor line, when for the ones further away from the line having the optical center unmodeled distortion errors would get progressively worse.

acpaquette:
This allows us to keep the new projective transform but takes into account more aggressive(?) distortion models. MRO CTX and LRONAC seem fairly tame compared to kaguyaTC but I am unfamiliar with what constitutes a more extreme distortion.

With the above information in mind, does this still seem good to merge?

oleg-alexandrov:
Looks good to me.

As to the initial guess projective transform, one could presumably compute it with pixels which are not image corners, like now, but more centrally located ones, where the distortion is less and this transform could be more accurate. (Note that the original linear approximation transform used corners too.) I don't think we should do it, that is an initial guess anyway, but maybe the distortion explains why that projective transform is not as good for kaguyaTC as it was for other sensors.

PR #437, opened 2023-03-24
First pass at projected line scan sensor model

acpaquette:
Adds in the initial CSM camera model to handle projected line scan camera images in USGSCSM.


src/Utilities.cpp
```diff
@@ -1528,3 +1528,45 @@ std::string ephemTimeToCalendarTime(double ephemTime) {
...
+
+std::string getProjectionString(json isd) {
+  std::string projection_string = "";
+  try {
+    projection_string = isd.at("proj_string");
+  } catch (...) {
+    throw std::runtime_error("Could not parse the projection string.");
+  }
+  return projection_string;
+}
```
Kelvinrr, line 1552: I imagine we would need to remove these in favor of the ALE ones?

Kelvinrr (commented):
just a  comment for now

CMakeLists.txt
```diff
@@ -75,11 +80,11 @@ set_target_properties(usgscsm PROPERTIES
     VERSION ${PROJECT_VERSION}
     SOVERSION 1
 )
-
 set(USGSCSM_INCLUDE_DIRS "${CMAKE_CURRENT_SOURCE_DIR}/include/usgscsm"
                          "${CMAKE_CURRENT_SOURCE_DIR}/include"
                          "${CMAKE_CURRENT_SOURCE_DIR}"
-                         "${EIGEN3_INCLUDE_DIR}")
+                         "${EIGEN3_INCLUDE_DIR}"
+                         "${PROJ_INCLUDE_DIR}")
```
Kelvinrr, line 87: do we need to setup gdal?

environment.yml
```diff
@@ -4,8 +4,9 @@ channels:
   - default
 
 dependencies:
-  - cmake>=3.15
   - ale>=0.8.8
+  - cmake>=3.15
   - csm
-  - nlohmann_json
   - eigen
+  - gdal
```
Kelvinrr, line 11: is gdal being used or can we get away with just proj?

CMakeLists.txt
acpaquette: Nope, ALE handles gdal and usgscsm only needs proj. The gdal in the environment file is a hold over

environment.yml
acpaquette: Nope, it can go

src/UsgsAstroProjectedLsSensorModel.cpp
```diff
@@ -0,0 +1,799 @@
...
+
+//***************************************************************************
+// UsgsAstroLineScannerSensorModel::reset
+//***************************************************************************
+void UsgsAstroProjectedLsSensorModel::reset() {
+  MESSAGE_LOG(spdlog::level::debug, "Running reset()");
+
+  m_geoTransform = std::vector<double>(6, 0.0);
+  m_projString = "";
+}
```
Kelvinrr, line 210: Should we call the parent reset?

src/UsgsAstroProjectedLsSensorModel.cpp
```diff
@@ -0,0 +1,799 @@
...
+// UsgsAstroLsSensorModel::groundToImage
+//***************************************************************************
+csm::ImageCoord UsgsAstroProjectedLsSensorModel::groundToImage(
+    const csm::EcefCoord &ground_pt, double desired_precision,
+    double *achieved_precision, csm::WarningList *warnings) const {
+  
+  PJ_CONTEXT *C = proj_context_create();
+
+  /* Create a projection. */
+  PJ *isdProj = proj_create(C, (m_projString + " +type=crs").c_str());
```
Kelvinrr, line 242: what is the  +crs for (sorry if you probably explained this already)

src/UsgsAstroProjectedLsSensorModel.cpp
```diff
@@ -0,0 +1,799 @@
...
+      spdlog::level::info,
+      "Meters {}, {}",
+      c_out.xyz.x, c_out.xyz.y);
+  std::vector<double> lineSampleCoord = meterToPixel(c_out.xyz.x, c_out.xyz.y, m_geoTransform);
+  csm::ImageCoord imagePt(lineSampleCoord[0], lineSampleCoord[1]);
+  MESSAGE_LOG(
+      spdlog::level::info,
+      "groundToImage result of ({}, {})",
+      imagePt.line, imagePt.samp);
+  return imagePt;
```
Kelvinrr, line 283: Seems like simply using proj was accurate enough for this without needing the camera model or image to ground?

src/UsgsAstroProjectedLsSensorModel.cpp
```diff
@@ -0,0 +1,799 @@
...
+//***************************************************************************
+csm::EcefCoordCovar UsgsAstroProjectedLsSensorModel::imageToGround(
+    const csm::ImageCoordCovar& image_pt, double height, double heightVariance,
+    double desired_precision, double* achieved_precision,
+    csm::WarningList* warnings) const {
+  csm::EcefCoord groundCoord = imageToGround(image_pt, height);
+  csm::ImageCoord cameraImagePt = UsgsAstroLsSensorModel::groundToImage(groundCoord);
+  csm::ImageCoordCovar cameraImagePtCovar(cameraImagePt.line, cameraImagePt.samp);
+  
+  return UsgsAstroLsSensorModel::imageToGround(cameraImagePtCovar, height, heightVariance, desired_precision, achieved_precision, warnings);
```
Kelvinrr, line 364: How is performance when having to go back and forth between image and ground?

src/UsgsAstroProjectedLsSensorModel.cpp
```diff
@@ -0,0 +1,799 @@
...
+// UsgsAstroProjectedLsSensorModel::imageToProximateImagingLocus
+//***************************************************************************
+csm::EcefLocus UsgsAstroProjectedLsSensorModel::imageToProximateImagingLocus(
+    const csm::ImageCoord& image_pt, const csm::EcefCoord& ground_pt,
+    double desired_precision, double* achieved_precision,
+    csm::WarningList* warnings) const {
+  csm::EcefCoord projGround = imageToGround(image_pt, 0);
+  csm::ImageCoord cameraImagePt = UsgsAstroLsSensorModel::groundToImage(projGround);
+  // std::cout.precision(17);
+  // std::cout << cameraImagePt.line << ", " <<
```
Kelvinrr, line 377: probably can delete these

src/UsgsAstroProjectedLsSensorModel.cpp
```diff
@@ -0,0 +1,799 @@
...
+  // Image id should include the suffix without the path name
+  m_imageIdentifier = imageId;
+}
+
+//***************************************************************************
+// UsgsAstroProjectedLsSensorModel::getSensorIdentifier
+//***************************************************************************
+std::string UsgsAstroProjectedLsSensorModel::getSensorIdentifier() const {
+  return m_sensorIdentifier;
+}
```
Kelvinrr, line 540: Im guessing a lot of the getters are not inherited?

Kelvinrr (commented):
I think with some simple tests comparing against the LS model somehow, this is good.

src/UsgsAstroProjectedLsSensorModel.cpp
acpaquette: No they are, I just didn't remove them, they are gone now

src/UsgsAstroProjectedLsSensorModel.cpp
acpaquette: No idea, fast enough that working with the camera in a jupyter notebook is not noticeable

src/UsgsAstroProjectedLsSensorModel.cpp
acpaquette: The parent class calls its reset in `UsgsAstroLsSensorModel::replaceModelState`, so we don't need to call it again

src/UsgsAstroProjectedLsSensorModel.cpp
acpaquette: To construct a proj string, you just have to tell proj that the proj4 string is a crs. Kinda weird but it errors without it

src/UsgsAstroProjectedLsSensorModel.cpp
acpaquette: Yes(?) I don't think involving the camera model would make a difference. We are in ground, so we could go back to camera image, then back to ground but you would just get the same ground. So you just feed it through the projection to get back to the projected pixel

Kelvinrr (approved):

PR #438, opened 2023-04-06
Generic Projected Camera

acpaquette:
Converts original LineScan projected subclass, to a generic projected CSM camera that holds the underlying USGSCSM model as a member variable.


Kelvinrr (approved):
Assuming the API is full covered, looks good.

PR #439, opened 2023-04-06
Small Projected Camera Fix

acpaquette:
A known .5 pixel offset exists between ISIS and CSM. This was being handled inside of the projected camera model, and should now be handled externally by the user if they are comparing ISIS and CSM


Kelvinrr:
Im guessing this doesn't fix the half pixel offset just updates the test data?

acpaquette:
@Kelvinrr It technically could but I think we want to maintain that .5 pixel offset

Kelvinrr (approved):

PR #445, opened 2023-08-08
USGSCSM Internal Dependency Building

acpaquette:
Updates USGSCSM to build everything internally

This PR integrates ALE and PROJ into USGSCSM and only building the parts that USGSCSM needs


environment.yml
```diff
@@ -4,9 +4,5 @@ channels:
   - default
 
 dependencies:
-  - ale>=0.8.8
   - cmake>=3.15
-  - csm
-  - eigen
-  - nlohmann_json
-  - proj
+  - sqlite
```
jlaura, line 8: This is concerning because many packages depend upon sqlite. For example:

- [ASP](https://github.com/NeoGeographyToolkit/StereoPipeline/blob/master/conda/asp_3.2.0_linux_env.yaml#L232)
- [GDAL](https://github.com/conda-forge/gdal-feedstock/blob/main/recipe/meta.yaml)
- SocetGXP???

Does the usgscsm depend on sqlite or is this because of the projection stuff inside the CSM?

thareUSGS: I think our last test build also needed the **curl** lib, unless that was removed from the build. 
https://github.com/OSGeo/PROJ/blob/a654e42074a167a03e1e3ff7c2009c04502830bf/docs/source/install.rst#build-requirements 

I can currently work-around the extra libs for SocetGXP, but it is not graceful. Ideally, bundling proj with usgscsm can be maybe optional (at least for Windows) but I'm not sure how best to resolve this.

recipe/meta.yaml
```diff
@@ -1,30 +1,39 @@
...
+{% set name = "usgscsm" %}
+{% set version = "main" %}
+
 package:
-  name: usgscsm
-  version: {{ environ.get("GIT_DESCRIBE_TAG", "") }}
+  name: {{ name|lower }}
+  version: {{ version }}
 
 source:
```
Kelvinrr, line 9: I imagine we would also need a PR to the feedstock to update the recipe?

CMakeLists.txt
```diff
@@ -42,53 +63,626 @@ if(USGSCSM_EXTERNAL_DEPS)
...
+  set(ALE_SOURCE_FILES ${CMAKE_CURRENT_SOURCE_DIR}/ale/src/InterpUtils.cpp
+                       ${CMAKE_CURRENT_SOURCE_DIR}/ale/src/Rotation.cpp
+                       ${CMAKE_CURRENT_SOURCE_DIR}/ale/src/Orientations.cpp
+                       ${CMAKE_CURRENT_SOURCE_DIR}/ale/src/States.cpp
+                       ${CMAKE_CURRENT_SOURCE_DIR}/ale/src/Util.cpp
+                       ${CMAKE_CURRENT_SOURCE_DIR}/ale/src/Vectors.cpp)
+
+  # set(ALE_HEADER_FILES ${ALE_BUILD_INCLUDE_DIR}/Rotation.h
+  #                      ${ALE_BUILD_INCLUDE_DIR}/Util.h)
+
```
Kelvinrr, line 649: Why is this commented out?

recipe/meta.yaml
acpaquette: Yes, this was just for internal testing that is similar to the feedstock. That's the next thing I will have to tackle once I pull a new release

CMakeLists.txt
acpaquette: It was a test to see if we needed the headers. I don't think the `ALE_HEADERS_FILES` variable is being used anywhere

acpaquette: I will remove the commented lines

Kelvinrr (approved):

PR #448, opened 2023-08-29
Doc updates and deploy

acpaquette:
Updates the docs to be a little more cohesive, also adds the job to deploy to AWS when merged in main


.github/workflows/aws_doc_deploy.yml
```diff
@@ -0,0 +1,37 @@
+name: aws-doc-deploy
+
+on:
+  workflow_run:
+    workflows: ["Build-Docs"]
```
jlaura, line 5: This is pretty sweet. It uses the name from any other run CI step as the key and if that completes, it fires? Nice.

.github/workflows/aws_doc_deploy.yml
```diff
@@ -0,0 +1,37 @@
...
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: us-west-2
+      
+      - name: Upload to S3
+        working-directory: docs/public
+        run: |
+          cd docs
+          aws s3 sync --delete public s3://asc-docs/usgscsm
```
jlaura, line 37: One thing that might make sense is to use a sub "directory" here. Like s3://asc-docs/software-docs or something? I am thinking of the use case where /tutorials has a site with tutorials, etc.

docs/source/user/isd_generation.rst
```diff
@@ -1,19 +1,21 @@
...
 Creating and testing an ISD
-===
+===========================
 
 This tutorial assumes that the user has created a conda environment with ale, usgscsm,
 and csmapi installed.
 
 Generating an ISD with ALE
------------------
+--------------------------
```
jlaura, line 10: I ❤️ sphinx and it's wonky warnings....

jlaura (approved):

.github/workflows/aws_doc_deploy.yml
acpaquette: Something to consider as the provided documentation matures!

.github/workflows/aws_doc_deploy.yml
acpaquette: Ideally, we will see when it gets merged

PR #455, opened 2023-09-11
FrameSensor: Make many members public so they can be changed

oleg-alexandrov:
The FrameSensor class has its members private, but has no set or get functions so those cannot be accessed. This is a proposal to make them public, as done already for the Linescan class. (A few members will still be kept private, which are internal states, such as the logger and the _state variable). 

I care primarily for the intrinsics (distortion, focal length), camera position, and camera orientation, as those are needed in bundle adjustment. But I don't think there's harm in having other class parameters be public, such as semi-major axis, semi-minor axis, etc. 

Thoughts?

### DIFF
--- a/include/usgscsm/UsgsAstroFrameSensorModel.h
+++ b/include/usgscsm/UsgsAstroFrameSensorModel.h
@@ -347,7 +347,6 @@ class UsgsAstroFrameSensorModel : public csm::RasterGM,
 
   static const std::string _SENSOR_MODEL_NAME;
 
- private:
   // Input parameters
   static const int m_numParameters;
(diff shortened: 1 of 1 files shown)


Kelvinrr:
I personally have no objection to making what would essentially be ISD components public. Makes sense to be able to modify those as needed directly without having to add setters and getters or similar.

oleg-alexandrov:
Any further thoughts on this?

Kelvinrr (approved):

Kelvinrr:
Nah I think this is fine if no one else has objections

oleg-alexandrov:
Can this be merged in?

PR #464, opened 2023-11-03
No need for tolerance check for radial distortion

oleg-alexandrov:
This is a proposed fix to https://github.com/DOI-USGS/usgscsm/issues/461.

The issue here is that with the unmodified code the radial distortion and undistortion operations are not being applied when the pixel value is very close to the origin. 

Ironically, this is when distortion and undistortion is most likely to converge well, as, radial distortion this is just some multiplications and additions. Radial undistortion is about the same thing.

There is no division by a number close to zero anywhere where one should need to ensure a tolerance must be respected.

I modified the unit test to make it more realistic, by using non-zero distortion coefficients. Before the proposed fix, the test will fail, as distortion and undistortion are not true inverse of each other. That is because, before the fix, the distortion operation was applied, since the pixel value is above the tolerance, but the undistortion operation is not applied, since then the pixel value is below the tolerance.

With this fix, the distortion and undistortion are consistently applied close to the origin and this unit test passes.

## Licensing

- [ x] I dedicate any and all copyright interest in this software to the public domain. I make this dedication for the benefit of the public at large and to the detriment of my heirs and successors. I intend this dedication to be an overt act of relinquishment in perpetuity of all present and future rights to this software under copyright law.

### DIFF
--- a/src/Distortion.cpp
+++ b/src/Distortion.cpp
@@ -94,11 +94,11 @@ void removeDistortion(double dx, double dy, double &ux, double &uy,
     case RADIAL: {
       double rr = dx * dx + dy * dy;
 
-        double dr = opticalDistCoeffs[0] +
-                    (rr * (opticalDistCoeffs[1] + rr * opticalDistCoeffs[2]));
+      double dr = opticalDistCoeffs[0] +
+                  (rr * (opticalDistCoeffs[1] + rr * opticalDistCoeffs[2]));
 
-        ux = dx * (1.0 - dr);
-        uy = dy * (1.0 - dr);
+      ux = dx * (1.0 - dr);
+      uy = dy * (1.0 - dr);
     } break;
 
     // Computes undistorted focal plane (x,y) coordinates given a distorted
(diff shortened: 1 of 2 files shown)


src/Distortion.cpp
```diff
@@ -327,14 +325,13 @@ void applyDistortion(double ux, double uy, double &dx, double &dy,
...
+    // focal plane coordinates. This case works by iteratively adding distortion
     // until the new distorted point, r, undistorts to within a tolerance of the
     // original point, rp.
     case RADIAL: {
       double rp2 = (ux * ux) + (uy * uy);
 
-      if (rp2 > tolerance) {
         double rp = sqrt(rp2);
         // Compute first fractional distortion using rp
         double drOverR =
```
acpaquette, line 337: I can't highlight the whole thing but can we tab in from lines 335 to 371 since it is no longer in the if statement?

src/Distortion.cpp
```diff
@@ -94,13 +94,11 @@ void removeDistortion(double dx, double dy, double &ux, double &uy,
     case RADIAL: {
       double rr = dx * dx + dy * dy;
 
-      if (rr > tolerance) {
         double dr = opticalDistCoeffs[0] +
                     (rr * (opticalDistCoeffs[1] + rr * opticalDistCoeffs[2]));
 
         ux = dx * (1.0 - dr);
         uy = dy * (1.0 - dr);
```
acpaquette, line 101: Can this be tabbed in?

tests/DistortionTests.cpp
```diff
@@ -124,17 +124,26 @@ TEST(transverse, removeDistortion_AlternatingOnes) {
   EXPECT_NEAR(uy, 7.5, 1e-8);
 }
 
-TEST(Radial, testRemoveDistortion) {
-  csm::ImageCoord imagePt(0.0, 4.0);
+TEST(Radial, testUndistortDistort) {
```
acpaquette, line 127: I'm a little concerned about this test change but it's likely fine(?)

acpaquette (commented):
Overall looks good, a couple formatting changes need and one potential concern from test changes

tests/DistortionTests.cpp
oleg-alexandrov: I will tab those in a bit. I left the text with original indentation to make it easier to see the diff. (Sloppy, yes.) 

Why is the distortion test a bit concerning? I on purpose chose a value to make the test fail before, because it was illustrating an inconsistency. I can easily go back to the previous pixel value, though likely any value will work as good as any other value, I am guessing.

I can also push down the tolerance value there if needed.

oleg-alexandrov: The code was not doing either distortion or understortion if the norm of the pixel was less than tolerance.

If the pixel is a little above tolerance, then distortion gets applied. The produced output pixel ends up a little below tolerance. Then undistortion is not applied.

This is a little silly, but a concrete example will help. Say x = 0.012, and tol = 0.01, and according to the formula distort(x) = 0.009. Then, normally, undistort(0.009) has to give back 0.012, but since 0.009 < tol, one gets undistort(0.009) = 0.009, which is the wrong answer. Granted, normally tol is small, so this does not matter much.

oleg-alexandrov: I fixed the indentation.

acpaquette (approved):

PR #466, opened 2023-11-07
Support for the radial + tangential distortion model

oleg-alexandrov:
Add support for the radial + tangential distortion model. This has 5 distortion coefficients, stored in order k1, k2, p1, p2, k3. The k1, k2, k3 coefficients are for radial distortion, and p1 and p2 for tangential distortion. This was validated with the OpenCV implementation using the function cv::projectPoints().

Corresponding code was added in "ale" (https://github.com/DOI-USGS/ale/pull/575). 

The issue discussing this: https://github.com/DOI-USGS/usgscsm/issues/465

Tests were added for:

 - Ensuring distortion and undistortion are inverse to each other
 - Loading an ISD and running image-to-ground and ground-to-image operations, and saving a state file. The ISD was obtained from a DawnFC example by replacing the DawnFC distortion (which is only radial with one coefficient) by the larger set of distortion coefficients which just get appended to the existing coefficient.

## Licensing

### DIFF
--- a/include/usgscsm/Distortion.h
+++ b/include/usgscsm/Distortion.h
@@ -6,7 +6,9 @@
 #include <tuple>
 #include <vector>
 
-enum DistortionType { RADIAL, TRANSVERSE, KAGUYALISM, DAWNFC, LROLROCNAC, CAHVOR, RADTAN };
+// This should be synched with the enum in ale/Distortion.h
+enum DistortionType { RADIAL, TRANSVERSE, KAGUYALISM, DAWNFC, LROLROCNAC, CAHVOR, 
+                     LUNARORBITER, RADTAN };
 
 // Transverse distortion Jacobian
 void transverseDistortionJacobian(double x, double y, double *jacobian,
@@ -23,6 +25,7 @@ void removeDistortion(double dx, double dy, double &ux, double &uy,
 void applyDistortion(double ux, double uy, double &dx, double &dy,
                      std::vector<double> const& opticalDistCoeffs,
                      DistortionType distortionType,
-                     const double desiredPrecision = 1.0E-6,
+                     const double desiredPrecisio
+                     n = 1.0E-6,
                      const double tolerance = 1.0E-6);
 #endif  // INCLUDE_USGSCSM_DISTORTION_H_
--- a/include/usgscsm/Distortion.h
+++ b/include/usgscsm/Distortion.h
@@ -25,7 +25,6 @@ void removeDistortion(double dx, double dy, double &ux, double &uy,
 void applyDistortion(double ux, double uy, double &dx, double &dy,
                      std::vector<double> const& opticalDistCoeffs,
                      DistortionType distortionType,
-                     const double desiredPrecisio
-                     n = 1.0E-6,
+                     const double desiredPrecision = 1.0E-6,
                      const double tolerance = 1.0E-6);
 #endif  // INCLUDE_USGSCSM_DISTORTION_H_
--- a/include/usgscsm/Distortion.h
+++ b/include/usgscsm/Distortion.h
@@ -6,11 +6,11 @@
 #include <tuple>
 #include <vector>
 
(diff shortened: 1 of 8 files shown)


include/usgscsm/Utilities.h
```diff
@@ -65,6 +65,15 @@ void lagrangeInterp(const int &numTime, const double *valueArray,
...
 
+// Use the Newton-Raphson method undistort a pixel (dx, dy), producing (ux, uy).
+void newtonRaphson(double dx, double dy, double &ux, double &uy,
+                    std::vector<double> const& opticalDistCoeffs,
+                    DistortionType distortionType, const double tolerance,
+                    std::function<void(double, double, double &, double &,
+                                       std::vector<double> const&)> distortionFunction,
+                    std::function<void(double, double, double *, 
+                                       std::vector<double> const&)> distortionJacobian);
+
```
acpaquette, line 76: Somewhat of a nitpick but would it make more sense for this to be part of the `Distortion.[cpp|h]` files? I guess I don't understand why a distortion function would be put into utils. Can it be used for more than computing an undistorted pixel?

oleg-alexandrov: My reasoning was that this would be a general purpose algorithm for inverting a 2D to to 2D function, to complement the current Brent algorithm which is in here as well. 

Note that Utilities.cpp already has a lot of distortion-related logic, such as computeDistortedFocalPlaneCoordinates(), getDistortionModel(), getDistortionCoeffs().

So, the function I added has general-purpose root-finding logic, but its interface is tied to lens distortion.

I can easily move it back to Distortion.cpp where it was before if it makes more sense.

acpaquette: The getDistortionModel(), getDistortionCoeffs() makes some sense as other ISD get functions are also present in the Utilities files. I think this makes enough sense the `newtonRaphson` function can stay here

include/usgscsm/Distortion.h
```diff
@@ -6,22 +6,24 @@
...
+                      std::vector<double> const& opticalDistCoeffs,
                       DistortionType distortionType,
                       const double tolerance = 1.0E-6);
 
 void applyDistortion(double ux, double uy, double &dx, double &dy,
-                     const std::vector<double> opticalDistCoeffs,
+                     std::vector<double> const& opticalDistCoeffs,
                      DistortionType distortionType,
                      const double desiredPrecision = 1.0E-6,
                      const double tolerance = 1.0E-6);
```
acpaquette, line 29: I am a little concerned here as this is API breaking, know we haven't started enforcing that in USGSCSM but we likely should. @Kelvinrr @jlaura Thoughts?

oleg-alexandrov: Thoughts on this are very welcome. I can easily undo this interface change. I assumed it was fine because these are internal functions, not mean to be called outside of usgscsm I think.

oleg-alexandrov: I rebased to latest and fixed a conflict

oleg-alexandrov:
Any comments as to whether changing the name of an internal function is fine or it has to be kept consistent? The new name was motivated by the fact that there are now two distortion Jacobians, for transverse and radtan distortion.

BTW, I think the binary compatibility may have been broken just because the Distortion enum has one more field now.

Anyhow, this is the last issue for this pull request, so hoping to hear how to proceed.

include/usgscsm/Distortion.h
Kelvinrr: I dont think this counts as API breaking? iirc const-ness doesn't change if it's in before or after the type. And passing by reference wound't change how a user interacts with the function.

oleg-alexandrov:
@acpaquette: Any firm decision on this? No problem adjusting things either way, then it could be pulled in.

oleg-alexandrov:
@acpaquette @jlaura: Thoughts? Happy to conform what is considered existing policy or best practices, if a specific change is suggested. There's no good reason for this to be left in limbo as people move on and stuff gets forgotten.

acpaquette (approved):

PR #484, opened 2024-04-29
Fix for radial-tangential distortion

oleg-alexandrov:
The OpenCV radial-tangential distortion model I added a few months ago is useful on the ASP side and helps with combining data from multiple vendors (and sensor types) under the CSM umbrella.

I found there is a bug, however, in my implementation, that is due to insufficiently clear OpenCV documentation regarding lens distortion at: https://docs.opencv.org/4.x/dc/dbb/tutorial_py_calibration.html

That does not say that distorted and undistorted pixels must be normalized by focal length. I found this bit by looking at another doc page, at https://docs.opencv.org/4.x/d9/d0c/group__calib3d.html

I put a fix and validated that this agrees with OpenCV's implementation. Before, I had tested only the distortion/undistortion formula, rather than the full chain of world-to-pixel transform, that's why this bug went undetected.

The existing CSM lens distortion formulas do not normalize by focal length. So this fix required breaking the interface of the distort/undistort functions, as now they need to know about the focal length, to apply/undo this normalization. 

I fixed the tests that broke as result of this and updated the API as needed.

## Licensing

### DIFF
--- a/include/usgscsm/Distortion.h
+++ b/include/usgscsm/Distortion.h
@@ -19,11 +19,13 @@ void computeTransverseDistortion(double ux, double uy, double &dx, double &dy,
 
 void removeDistortion(double dx, double dy, double &ux, double &uy,
                       std::vector<double> const& opticalDistCoeffs,
+                      double focalLength,
                       DistortionType distortionType,
                       const double tolerance = 1.0E-6);
 
 void applyDistortion(double ux, double uy, double &dx, double &dy,
                      std::vector<double> const& opticalDistCoeffs,
+                     double focalLength,
                      DistortionType distortionType,
                      const double desiredPrecision = 1.0E-6,
                      const double tolerance = 1.0E-6);
--- a/src/Distortion.cpp
+++ b/src/Distortion.cpp
@@ -84,13 +84,14 @@ void computeTransverseDistortion(double ux, double uy, double &dx, double &dy,
   }
 }
 
-// Compute distorted focal plane coordinates given undistorted coordinates. Use
-// the radial-tangential distortion model with 5 coefficients (k1, k2, k3 for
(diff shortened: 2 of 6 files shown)


oleg-alexandrov:
Any thoughts? I use the radial-tangential distortion model quite a bit, and is likely the default go-to model for a generic sensor.

acpaquette:
@oleg-alexandrov appreciate the second ping, this got buried in my inbox. My immediate thought is could the focal length be put into the coefficients list for use within the distortion algorithm? I know it wouldn't be an actual coefficient but it would prevent the API from being broken. I am also not sure if this would constitute a breaking change, considering it is an internal function that is not exposed to users

jlaura:
In my experience pushing heterogeneous things into a vector significantly reduces the readability and accessibility of the code base. Isn't this a bug fix?

oleg-alexandrov:
The focal length could be made an optional last argument if need be, but I don't see a good reason for that. 

This is indeed an internal function, so changing the API should not break anything. 

There is a very good reason as to why a distortion function must know the focal length. This assures that the distortion is applied correctly independently of image resolution. If you put a new sensor of higher resolution inside of your camera, but keep the same lens, the distortion should function as before.

I understand your reluctance to break the API, but I don't think this is a concern for this particular function.

acpaquette:
All fair points, I guess it just seems weird that none of the other distortion algorithms for both applying and removing distortion care about the focal length. 

It makes a little more sense that a generic distortion model would.

oleg-alexandrov:
I think an orbital sensor is usually a very custom thing, with fixed resolution, to be calibrated as the chief scientist wishes. People running bots on the ground want the flexibility of swapping the "eyes" of their machines with predictable consequences.

acpaquette (approved):

PR #490, opened 2025-02-14
Cahvor fix

### LINKED ISSUE #488: Issues with lens distortion computation
This is similar to https://github.com/DOI-USGS/usgscsm/issues/461, but applies to the cahvor distortion model.

For cahvor distortion, the formula (https://github.com/DOI-USGS/usgscsm/blob/main/src/Distortion.cpp#L325) is:

     double rr = shiftedDx * shiftedDx + shiftedDy * shiftedDy;

      if (rr > tolerance)
      {
        double dr = opticalDistCoeffs[0] +
                    (rr * (opticalDistCoeffs[1] + rr * opticalDistCoeffs[2]));

        ux = shiftedDx * (1.0 - dr);
        uy = shiftedDy * (1.0 - dr);
        ux += opticalDistCoeffs[3];
        uy += opticalDistCoeffs[4];
      }
    }
    break;

When rr is smaller than tolerance, the undistortion ux and uy is assumed to be equal to input distorted values, which is wrong.

Need to add the optical shifts given by opticalDistCoeffs[3] and [4] even when rr is small.

This formula should not have an if statement at all. This is a simple polynomial formula, that works for any input.

The same for cahvor undistortion, at https://github.com/DOI-USGS/usgscsm/blob/main/src/Distortion.cpp#L585

There is no need to consider a tolerance. The only division by a value is:

 dx = shiftedUx / (1.0 - drOverR);
 dy = shiftedUy / (1.0 - drOverR);

and here it doesn't matter if we are close to the origin, as that does not make 1.0 - drOverR be equal to 0, rather, this quantity is close to 1.0.

AustinSanders:
Closes #488 

## Licensing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,9 @@ release.
 
 ## [Unreleased]
 
+### Fixed
+- Fixed CAHVOR model optical shifts by removing tolerance check [#488](https://github.com/DOI-USGS/usgscsm/issues/488)
+
 ## [2.0.1] - 2024-01-23
 
 ### Changed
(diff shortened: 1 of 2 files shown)


acpaquette (approved):
Nice, much cleaner and clearer!

PR #492, opened 2025-04-04
Frame cam no bounds

### LINKED ISSUE #489: Frame camera throws out-of-bounds exception
The frame camera model code throws an exception when it is asked to find the sensor position and velocity when the input pixel is a bit outside of the image domain. This is for functions UsgsAstroFrameSensorModel::getSensorPosition() and getSensorVelocity().

This is not desirable behavior. It is possible for a ground point to project into the camera at a pixel outside of the image box. The software should be able to return the sensor position and velocity for such a pixel, which is the same for all pixels anyway, for a frame camera.

This check is not present for linescan cameras, which graciously does the best it can. It clamps the input pixel line to the range of lines it can handle.

Such behavior as for the frame camera is problematic as it requires fixes in software that functions perfectly fine with other camera implementations, including outside of usgscsm.

jrcain-usgs:
- Removed any checks at all for pixel boundary in UsgsAstroFrameSensorModel::getSensorPosition(imageCoord).
- Added a test to check a pixel exceeding the boundary.

Fixes #489 

## Licensing

### DIFF
--- a/tests/FrameCameraTests.cpp
+++ b/tests/FrameCameraTests.cpp
@@ -533,7 +533,10 @@ TEST_F(FrameSensorModelLogging, GetSensorPositionPixel) {
 
 TEST_F(FrameSensorModelLogging, GetSensorPositionPixelExceedsBounds) {
   csm::ImageCoord imagePt(-1, -1);
-  sensorModel->getSensorPosition(imagePt);
+  csm::EcefCoord sensorPos = sensorModel->getSensorPosition(imagePt);
+  EXPECT_EQ(sensorPos.x, 1000);
+  EXPECT_EQ(sensorPos.y, 0);
+  EXPECT_EQ(sensorPos.z, 0);
 }
 
 TEST_F(FrameSensorModelLogging, GetSensorPositionTime) {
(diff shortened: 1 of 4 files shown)


chkim-usgs:
An update to the changelog and it should be good to go.

tests/FrameCameraTests.cpp
```diff
@@ -531,6 +531,11 @@ TEST_F(FrameSensorModelLogging, GetSensorPositionPixel) {
   sensorModel->getSensorPosition(imagePt);
 }
 
+TEST_F(FrameSensorModelLogging, GetSensorPositionPixelExceedsBounds) {
+  csm::ImageCoord imagePt(-1, -1);
+  sensorModel->getSensorPosition(imagePt);
```
acpaquette, line 536: A slightly better test would be to do a round robin using a pixel that is outside of the image. Basically, run `imageToGround`, get the ground coord, then run `groundToImage` and see what pixel the camera returns

acpaquette: After some discussion and digging, the framing camera model doesn't use `getSensorPosition` in the `imageToGround` call so this test can more or less stay as is. It would be a good idea to still check the x, y, z output from `getSensorPosition`

jrcain-usgs: Discussed this with @acpaquette, we decided it would actually be better just to add a check on the position.

chkim-usgs (approved):

PR #496, opened 2025-05-18
Fix for m_targetName not being carried along

oleg-alexandrov:
This is a minor fix for the Frame camera forgetting to keep track of the target name (planetary body).

Note that ALE does not have a member function for the target name. So, this is a very narrow fix only for the case when the frame camera was explicitly told what the target name is by other means and is expected to remember that when recreating a model from a state file.

Longer term, likely in ALE one should ensure the target name is read from either SPICE or ISIS cubes (the latter sets this), then passed along to USGSCSM, but that is beyond the scope of this fix.

How to test: Populate this field in a json file, then call usgscsm_cam_test to read and then write this back to a new file, as:

usgscsm_cam_test --model  input.json --output-model-state output.json
 
## Licensing


Kelvinrr:
What are you using the target name for? ISIS Cubes track the target name since they are tied to a particular SET (ISIS), but CSM models are not. Tracking targets seems more like the image formats responsibility. 

And if we need track the target here we can update all the models. I think `m_targetName` is mostly a holdover from an old implementation. I would be more inclined to remove it unless we determine CSM models should track all target info (including shape models). Which I am open to, I am just saying there are design implications.

oleg-alexandrov:
m_targetName is the only place where the user can put the planet name. 

An image does not know on what planet it was made, it is just a TIF usually (CSM supports non-cub images). 

The target name belongs with the camera metadata, together with planet radius, etc.

Updating all the models and tracking shape model is already a big extension of this. Not saying I am opposed to it, but for the moment this little change would go the right way. There are other fields in the json that are optional, and when they exist, they are caried over.

oleg-alexandrov:
In my current work with JunoCam, it is nice to distinguish between Ganymede, Europa, and Io flybys. The target name in CSM cameras is the only way to set that.

All satellite vendors store metadata in camera files, not image files. ISIS cub files are a bit of an outlier here.

I will strongly advocate support for target name in CSM, and this pull request is a small step in the right direction.

PR #498, opened 2026-01-21
fixes incorrect J2000 time

amystamile-usgs:
## Licensing


amystamile-usgs:
This doesn't completely fix the time in comparison to ISIS (still off by ~60s). I think the fix would have to be to include a leapseconds table. But this is closer to the correct time. The original code is off by 12 hours.

Kelvinrr:
closing in favor of #506

PR #503, opened 2026-04-17
remove JSON from API 

Kelvinrr:
Updated to remove JSON from API. Also added a VariantMap to be a simple map object that can be used in place of the json lib to remove the need for a external type but also reduce the amount of times the state is dumped to JSON or a str. 

## Licensing

- [ ] I dedicate any and all copyright interest in this software to the public domain. I make this dedication for the benefit of the public at large and to the detriment of my heirs and successors. I intend this dedication to be an overt act of relinquishment in perpetuity of all present and future rights to this software under copyright law.


include/usgscsm/UsgsAstroLsSensorModel.h
```diff
@@ -63,20 +63,18 @@ class UsgsAstroLsSensorModel : public csm::RasterGM,
...
+  void populateModel(const std::string& j);
+  std::string getModelJson() const;
 
   // This method checks to see if the model name is recognized
   // in the input state string.
   static std::string getModelNameFromModelState(const std::string& model_state);
 
-  std::string constructStateFromIsd(const std::string imageSupportData,
-                                    csm::WarningList* list);
+  VariantMap constructStateFromIsd(const std::string imageSupportData,
```
Kelvinrr, line 76: Thinking about it now, we should check this function is not part of the CSM API. 

EDIT: it is not

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -180,131 +179,125 @@ void UsgsAstroLsSensorModel::populateModel(const nlohmann::json& j) {
...
     createProjectiveApproximation();
   } catch (...) {
     m_useApproxInitTrans = false;
   }
 }
 
+void UsgsAstroLsSensorModel::populateModel(const std::string& stateString) {
+  populateModel(variantMapFromJson(stateAsJson(stateString)));
+}
+
```
acpaquette: Can this be moved before the other `populateModel` function? I know it doesn't matter to the computer but it makes the code a bit easier to read

include/usgscsm/UsgsAstroLsSensorModel.h
```diff
@@ -63,20 +63,18 @@ class UsgsAstroLsSensorModel : public csm::RasterGM,
...
-  // serialization. Used by replaceModelState() and binary state loading.
-  void populateModel(const nlohmann::json& j);
+  void populateModel(const VariantMap& vm);
+  VariantMap getModelMap() const;
 
-  // Return model state as a json object, bypassing string serialization.
-  // Used by getModelState() and binary state saving.
-  nlohmann::json getModelJson() const;
+  void populateModel(const std::string& j);
+  std::string getModelJson() const;
```
acpaquette: I'm not sure if we want to keep these, seems like the VarientMap should just replace these json functions. If we want to retain the function signature for `constructStateFromIsd` then I can see keeping these functions

include/usgscsm/UsgsAstroFrameSensorModel.h
```diff
@@ -38,8 +37,8 @@ class UsgsAstroFrameSensorModel : public csm::RasterGM,
 
   static std::string getModelNameFromModelState(const std::string& model_state);
 
-  std::string constructStateFromIsd(const std::string &jsonIsd,
-                                    csm::WarningList *warnings);
+  VariantMap constructStateFromIsd(const std::string &jsonIsd,
+                                   csm::WarningList *warnings);
```
acpaquette, line 41: Should this be considered an API break? Do we only care about the virtual functions from `RasterGM` and `Ellipsoid`?

include/usgscsm/Utilities.h
```diff
@@ -26,15 +27,15 @@ void computeDistortedFocalPlaneCoordinates(
 
 void removeJitter(
     const double &line, const double &sample,
-    const std::vector<double> lineJitterCoeffs, const std::vector<double> sampleJitterCoeffs,
-    const std::vector<double> lineTimes,
+    const std::vector<double>& lineJitterCoeffs, const std::vector<double>& sampleJitterCoeffs,
+    const std::vector<double>& lineTimes,
```
acpaquette, line 31: Reference for speed?

acpaquette (commented):
Cpp file change seem fine. Largely just need to determine if these changes are API breaking

include/usgscsm/UsgsAstroFrameSensorModel.h
Kelvinrr: I dont think so since these functions are not part of the CSM API technically? afaik we mostly have to worry about maintaining the CSM API. We would need to check if it's being used directly in ISIS though.

include/usgscsm/Utilities.h
Kelvinrr: yeah, there were a lot of places copying large arrays.

src/UsgsAstroLsSensorModel.cpp
Kelvinrr: removed them

src/UsgsAstroFrameSensorModel.cpp
```diff
@@ -1834,13 +1922,13 @@ std::string UsgsAstroFrameSensorModel::constructStateFromIsd(
...
   state["m_referencePointXyz"] = std::vector<double>(3, 0.0);
-  state["m_currentParameterCovariance"] =
-      std::vector<double>(NUM_PARAMETERS * NUM_PARAMETERS, 0.0);
+  state["m_currentParameterCovariance"] = std::vector<double>(NUM_PARAMETERS * NUM_PARAMETERS, 0.0);
   for (int i = 0; i < NUM_PARAMETERS * NUM_PARAMETERS;
        i += NUM_PARAMETERS + 1) {
     state["m_currentParameterCovariance"][i] = 1;
   }
   state["m_collectionIdentifier"] = "";
+  std::cout << "set the covariance to identity" << std::endl;
```
acpaquette: Remove debug print

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -126,52 +126,40 @@ const csm::param::Type UsgsAstroLsSensorModel::PARAM_CHAR_ALL[] = {
...
- * @brief Replaces the sensor model's state with a new state.
- * @description This function updates the internal state of the sensor model based on a 
- * given state string. The state string is parsed, and its contents are used to update
- * model parameters. This function is primarily used for sensor model instantiation or 
- * updating the model state in iterative optimization processes.
- * 
- * @param stateString A JSON string representing the new state of the sensor model. This 
- * string should contain all necessary model parameters and metadata to fully define the 
- * sensor model state.
- */
```
acpaquette, line 139: Why was this removed?

src/UsgsAstroLsSensorModel.cpp
```diff
@@ -180,131 +168,122 @@ void UsgsAstroLsSensorModel::populateModel(const nlohmann::json& j) {
...
     }
   }
 
-  // If computed state values are still default, then compute them
   if (m_gsd == 1.0 && m_flyingHeight == 1000.0) {
     updateState();
   }
 
   try {
-    // Approximate the ground-to-image function with a projective transform
```
acpaquette, line 301: Removing these comments seems unnecessary, can this change be reverted?

include/usgscsm/UsgsAstroPluginSupport.h
```diff
@@ -3,11 +3,12 @@
...
-#include <nlohmann/json_fwd.hpp> // forward declaration
+#include "VariantMap.h"
 
 csm::RasterGM *getUsgsCsmModelFromIsd(const std::string &stringIsd, const std::string &modelName, csm::WarningList *warnings);
-csm::RasterGM *getUsgsCsmModelFromState(const std::string &stringState, const std::string &modelName, csm::WarningList *warnings);
-csm::RasterGM *getUsgsCsmModelFromJson(const nlohmann::json &j, const std::string &modelName, csm::WarningList *warnings);
-nlohmann::json getUsgsCsmModelJson(csm::RasterGM *model);
+csm::RasterGM *getUsgsCsmModelFromJsonState(const std::string &jstring, const std::string &modelName, csm::WarningList *warnings);
+csm::RasterGM *getUsgsCsmModelFromVariantMap(const VariantMap &vm, const std::string &modelName, csm::WarningList *warnings);
+std::string getUsgsCsmModelJson(csm::RasterGM *model);
```
acpaquette, line 11: Again, potential API breakage

src/Utilities.cpp
```diff
@@ -2659,4 +2665,63 @@ std::vector<double> meterToPixel(double meter_x, double meter_y, std::vector<dou
   double line = (meter_y - geoTransform[3]) / geoTransform[5];
 
   return {line, sample};
+}
+
+VariantMap variantMapFromJson(const nlohmann::json& j) {
```
acpaquette, line 2670: It seems like we have just shifted our JSON API issues to the Utilities. Shouldn't this take a string rather than a json object?

src/Utilities.cpp
```diff
@@ -2659,4 +2665,63 @@ std::vector<double> meterToPixel(double meter_x, double meter_y, std::vector<dou
...
+      } else if ((*it)[0].is_number()) {
+        result.set<std::vector<double>>(key, it->get<std::vector<double>>());
+      }
+    }
+  }
+
+  return result;
+}
+
+nlohmann::json jsonFromVariantMap(const VariantMap& vm) {
```
acpaquette, line 2698: Similar here, shouldn't this return a string rather than json?

include/usgscsm/UsgsAstroPushFrameSensorModel.h
```diff
@@ -64,19 +64,18 @@ class UsgsAstroPushFrameSensorModel : public csm::RasterGM,
...
 
-  // Populate model fields directly from a json object, bypassing string
-  // serialization. Used by replaceModelState() and binary state loading.
-  void populateModel(const nlohmann::json& j);
+  void populateModel(const VariantMap& vm);
+  VariantMap getModelMap() const;
 
-  // Return model state as a json object, bypassing string serialization.
-  nlohmann::json getModelJson() const;
+  void populateModel(const std::string& j);
```
acpaquette: Missed one of the populateModel functions

acpaquette (commented):
Couple more things to address

src/Utilities.cpp
Kelvinrr: There is a ton of stuff utilities that use JSON directly, I think it's mostly fine since these are compiled into the library and are not used directly in the cameras API which is where you get into trouble.

src/Utilities.cpp
Kelvinrr: I originally used strings but this caused a lot of re-parsing. I can double check the work flows since I think I removed most of the areas that were re-parsing in the first place when creating a camera.

include/usgscsm/UsgsAstroPluginSupport.h
Kelvinrr: I can see how they're being used in ISIS

Kelvinrr: It's not being used

include/usgscsm/UsgsAstroPushFrameSensorModel.h
Kelvinrr: forgot to git add

src/UsgsAstroLsSensorModel.cpp
Kelvinrr: I think I originally deleted the function before bringing it back after realizing I still need it. Reverted the comment.

.github/workflows/ci_testing.yml
```diff
@@ -43,11 +43,11 @@ jobs:
           cd build
           export PROJ_DATA=$PWD/PROJ/data
           cmake -DCMAKE_BUILD_TYPE=RELEASE -DUSGSCSM_EXTERNAL_DEPS=OFF -DUSGSCSM_BUILD_DOCS=OFF ..
-          cmake --build .
+          cmake --build . -j $(nproc)
```
acpaquette: The following error/debug message is in the build logs. We should either fix it or remove the `-j $(proc)`
```
/Users/runner/work/_temp/34fa6760-20e7-4263-b0a6-5d27d946315a.sh: line 5: nproc: command not found
```

acpaquette (approved):

PR #506, opened 2026-05-28
Fix j2000 time

oleg-alexandrov:
This fixes a merge conflict in @amystamile-usgs's pull request https://github.com/DOI-USGS/usgscsm/pull/498 and adds a table of leap seconds to make the result agree precisely with NAIF SPICE.

This was coded up with Claude but it is honest work. There is a test to ensure the output of ephemTimeToCalendarTime agrees with known NAIF SPICE values.

If one runs the attached script [verify_ephemtime.py](https://github.com/user-attachments/files/28327279/verify_ephemtime.py) this will produce the NAIF SPICE reference values. The test values match NAIF exactly for all dates from 1972 onward, within 1 ms. The pre-1972 entry may differ by ~1 s because NAIF uses a continuous formula while we use the discrete leap-second table.

Looks like low-hanging fruit for a pull request that kind of fell behind.

## Licensing


oleg-alexandrov:
This looks like a small easy to merge bugfix.

Kelvinrr:
essentially re-implementing utc to et so avoid LSKs is not a bad idea,

Kelvinrr (approved):

PR #507, opened 2026-06-02
Wasm support

Kelvinrr:
Adds web assembly build support to USGSCSM. Excludes the projected sensor, but since working on this I figured out how to include the projected sensor. I can add it in a future PR. 

## Licensing


oleg-alexandrov:
Drive-by, uninvited review here, hope that is welcome. usgscsm is a dependency for ASP, so we would pick these changes up downstream, and I had Claude help me go through the diff carefully before writing this up.

Big-picture note first: the title says WASM, but the larger change in this PR is actually replacing the vendored spdlog with a hand-rolled Logging.h, which is what touches every sensor model. That is the part I would most want called out, because it changes runtime behavior and is not just a build addition.

A few specific things stood out.

1. There appear to be leftover debug prints in UsgsAstroPlugin::loadImageSupportData. Three std::cout lines ("Loading image support data from ISD...", "creating ifstream with filename...", "done") were added and fire on every model construction, writing to stdout. Those look accidental, and they would pollute stdout for any ISIS or ASP tool that loads a CSM model.

2. The logging default seems to have flipped from off to on. Before, logging was effectively silent unless USGSCSM_LOG_FILE was set (m_logger was null). The new Logging.h defaults to INFO and always writes to stderr, so every LOG_INFO now prints by default. For tools that construct many models that is a lot of new output on stderr. Related: USGSCSM_LOG_FILE no longer does anything, which is a silent backward-compatibility break for anyone routing logs to a file.

3. Smaller points on the new logger itself: std::localtime is not thread-safe, the log level is a non-atomic shared static, and the formatter only understands bare {} placeholders, so any spdlog-style spec such as {:.6f} or {0} would be printed literally rather than formatted. spdlog handled all of that already. It may be worth weighing this against simply keeping spdlog for the native build and only excluding it from the WASM target.

4. Would it be feasible to split this into two PRs, or at least describe the spdlog removal in the PR body? As it stands the +2.6k / -16.7k diff reads like cleanup, but the logging swap is a behavior change that deserves its own discussion.

The WASM side itself looks nicely contained. The ifndef __EMSCRIPTEN__ guards and the separate usgscsm_wasm target keep it from disturbing the native build, and excluding only the projected sensor to avoid the PROJ dependency is a sensible call. Nice work on that part.

Thanks for putting this together.

oleg-alexandrov:
Thanks for the quick turnaround, I went back through the diff (again with Claude's help). The main thing is addressed: the stray std::cout prints are gone, and flipping the default log level to OFF means the logger is silent unless explicitly enabled, so no more stdout/stderr pollution for tools that construct many models. The localtime_r/localtime_s swap is a nice cross-platform thread-safety fix too. From the ASP downstream side, this is good to go.

A few minor leftovers, none of which I would block on, just noting for the record:

1. current_log_level is still a non-atomic static. Harmless with the default OFF and single-threaded construction, only a concern if levels get toggled across threads.

2. USGSCSM_LOG_FILE no longer routes logs to a file (output goes to stderr now). With the default OFF the practical regression is gone, but it is a silent compat break for anyone who previously set that variable.

3. The formatter still only understands bare {} placeholders, so spdlog-style specs like {:.6f} or {0} would print literally.

All small. Merge whenever it makes sense to you. Nice work.

Kelvinrr:
Separating the PR between WASM and Logging turned out to be a nightmare, so I think we'll just keep them together since removing spdlog was required for WASM

oleg-alexandrov:
I agree.

PR #509, opened 2026-06-04
removed spdlog

Kelvinrr:
@acpaquette suggested I split removing SPDLOG from WASM, so this is that PR. It was easier to just redo it to mimic the spdlog API to minimize changes to the code. I'll update WASM PR to be just WASM. 

## Licensing


tests/Fixtures.h
acpaquette, line 66: This needs to be handled like in the WASM PR

PR #502, opened 2026-03-27
Add isUsgsCsmIsd() and isUsgsCsmState() quick peek functions

oleg-alexandrov:
The usgscsm_cam_test tool has a very complex way of loading a model. 

First, it iterates over all plugins.  For each, it tries to construct an isd which does a lot of work. If it succeeds it constructs the isd (so repeat of before). If not in luck, it tires to construct state for each plugin, and if in luck, it again constructs the state.

This can be very slow and I suspect similar logic with first try to construct and then actually construct (so 2x the work) is in other places too.

This adds quick functions that scan the model string (avoiding both json parsing and model building). One function is an isd check, and second is a model state check (that covers .sup as well). On success it returns the needed id that is later used for actual construction.

If this is accepted, these functions could likely be used in other places too.

## Licensing
This project is mostly composed of free and unencumbered software released into the public domain, and we are unlikely to accept contributions that are not also released into the public domain. Somewhere near the top of each file should have these words:

> This work is free and unencumbered software released into the public domain. In jurisdictions that recognize copyright laws, the author or authors of this software dedicate any and all copyright interest in the software to the public domain.

- [x] I dedicate any and all copyright interest in this software to the public domain. I make this dedication for the benefit of the public at large and to the detriment of my heirs and successors. I intend this dedication to be an overt act of relinquishment in perpetuity of all present and future rights to this software under copyright law.

PR #525, opened 2026-07-15
Attempt the projected sensor model only for projected ISDs

oleg-alexandrov:
## Summary

`constructModelFromISD` built a projected sensor model first for every ISD, then fell back to the unprojected model if that threw. The projected model needs a geotransform in the ISD. A frame ISD, and an unprojected linescan ISD, do not have one, so the projected attempt always threw for them and the failure was logged at error level, even though the correct model was then built. The result is a misleading error printed on every normal load. For example, loading a frame ISD such as tests/data/simpleFramerISD.json prints:

```
[USGSCSM][ERROR][constructModelFromISD] Could not construct model [USGS_ASTRO_FRAME_SENSOR_MODEL] with error [Could not parse the geo_transform. ERROR: key 'geotransform' not found]
```

A valid model is still returned, so nothing is actually wrong; the error is noise.

## Fix

Attempt the projected model only when the ISD declares a projection, detected by the presence of a geotransform, following the quick string-scan style of `isUsgsCsmIsd` from #502. A non-projected ISD now goes straight to the unprojected model with no attempt and no message. A projected ISD that genuinely fails is still logged at error level, so real errors are not hidden. This is a follow-up to #437, which introduced the projected sensor model and the try-then-fallback path.

## How to reproduce

Load any non-projected ISD, for example the frame ISD tests/data/simpleFramerISD.json, through the plugin or usgscsm_cam_test. Before this change an error-level line about a missing geotransform is printed although the model loads fine. After this change the load is quiet. Projected ISDs, such as tests/data/constVelocityProj.json, are unaffected.

## Testing

Added a LoggingTests regression that constructs a model from tests/data/simpleFramerISD.json and asserts that no error-level line is logged and that a model is returned.

This was done with Claude (AI) assistance.

## Licensing


oleg-alexandrov:
To add I got report of this from another user as well. It reads like a failure and people wonder what to do.

src/UsgsAstroPlugin.cpp
```diff
@@ -492,25 +492,29 @@ csm::Model *UsgsAstroPlugin::constructModelFromISD(
...
+      return projModel;
+    } catch (std::exception &e) {
+      delete projModel;
+      std::string aMessage = "Could not construct model [";
+      aMessage += modelName;
+      aMessage += "] with error [";
+      aMessage += e.what();
+      aMessage += "]";
+      LOG_ERROR( aMessage);
+    }
```
acpaquette, line 526: This should likely throw an exception rather than just an error log message to avoid trying to build the unprojected sensor

acpaquette (commented):
Overall looks good! One small thing to address

src/UsgsAstroPlugin.cpp
oleg-alexandrov: Done. When the ISD declares a projection but the projected model fails, it now logs the error and throws a csm::Error of type SENSOR_MODEL_NOT_CONSTRUCTIBLE instead of falling through to the unprojected model. It logs and throws, rather than throwing alone, so the message reaches the usgscsm log regardless of what the caller does, while the throw stops the fall-through and passes the real reason upstream. This follows the log-then-throw pattern of constructModelFromState in the same file. Prepared with Claude (AI) assistance.

acpaquette (approved):

PR #528, opened 2026-07-20
Adds support for cloud optimized state files

Kelvinrr:
Adds support for cloud optimized state files via the [StarDS file format](https://doi-usgs.github.io/StarDS/). 

This allows users to open Camera models over HTTP and S3. Locally shows massive performence increase when create CTX models (~0.02s StarDS vs ~2s JSON per camera), instiantiating over HTTP was about as fast as JSON if not a little bit faster on my informal benchmarks. 

This also redoes how USGSCSM handles external deps. This is partly since we need to include more deps for remote access and compression. But also, we need to ship USGSCSM without external deps. This bakes in PROJ with the DB into the library so there should be 0 external deps at runtime. 

## Licensing

- [ ] I dedicate any and all copyright interest in this software to the public domain. I make this dedication for the benefit of the public at large and to the detriment of my heirs and successors. I intend this dedication to be an overt act of relinquishment in perpetuity of all present and future rights to this software under copyright law.


CMakeLists.txt
```diff
@@ -49,6 +57,11 @@ option(ENABLE_TIFF "Set to build the TIFF components of proj" OFF)
 option(BUILD_TESTING "Set to build the proj tests" OFF)
 option(BUILD_APPS "Set to build the proj apps" OFF)
 
+# STARDS state-file input support. Off by default: enabling it compiles the
```
acpaquette: `On` by default?

CMakeLists.txt
```diff
@@ -97,25 +123,60 @@ if(USGSCSM_EXTERNAL_DEPS)
 else()
   set(BUILD_SHARED_LIBS OFF)
   message(STATUS "Using embedded dependencies")
-
-  # Skip PROJ library for WebAssembly builds (projected sensor model is disabled)
-  if(NOT EMSCRIPTEN)
+  # CSM API library
+  add_subdirectory(external/csm)
```
acpaquette: Can we check this with the CSM folks? We basically need to know if no linking includes the CSM api library, it seems weird to me that the plugin itself would need to build the CSM api into itself

CMakeLists.txt
```diff
@@ -97,25 +123,60 @@ if(USGSCSM_EXTERNAL_DEPS)
...
+  # Directory holding the CSM headers (RasterGM.h, Error.h, ...). Consumers that
+  # include them bare (e.g. #include <RasterGM.h>) need this on their path.
+  set(CSM_INCLUDE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/external/csm")
+  set(CSM_LIBRARY csmapi)
+
+  # PROJ (for the projected sensor model) is built for both native and WASM.
+  # Its database, proj.db, is embedded and served from memory via a custom
+  # SQLite VFS, so no runtime file or PROJ_DATA is needed on either platform.
+  if(TRUE)
```
acpaquette: Can this `if` just be removed?

CMakeLists.txt
```diff
@@ -638,41 +733,59 @@ else()
...
+  #   * ale/Distortion.h: conda's older copy lacks newer enum values (e.g.
+  #     KPLOSHADOWCAM), breaking the tests.
+  #   * nlohmann/json.hpp: conda ships a 3.11.2 built WITHOUT the CustomBaseClass
+  #     template parameter, while external/json (also 3.11.2) has it. Both stamp
+  #     the same json_abi_v3_11_2 inline namespace, so basic_json mangles to
+  #     different symbols (extra trailing 'v'/void arg) depending on which header
+  #     a TU sees. The library (Utilities.cpp) compiles against external/json but
+  #     the test TUs were picking up conda's, producing undefined-symbol link
+  #     errors for every ISD getter. BEFORE forces the vendored copies first on
+  #     every target's include path.
```
acpaquette: This is a bit to technical for a comment, can it be reduced?

CMakeLists.txt
```diff
@@ -718,31 +831,100 @@ target_include_directories(usgscsm
...
-                        ${PROJ_TARGET})
+# STARDS support: compile the header-only STARDS reader (from the external/StarDS
+# submodule) into usgscsm. We do NOT use the submodule's own CMakeLists (it builds
+# tools/tests/SWIG bindings and resolves its compression/HTTP deps via
+# find_package to dynamic system libs). Instead we consume just its header via our
+# own INTERFACE target and supply the dependencies the way the usgscsm plugin
+# needs: vendored static libs (no runtime dylibs) natively, and the emscripten
+# zlib port for WASM. This target carries the STARDS feature defines plus the
+# dependency include dirs and link flags/libraries, propagated to consumers.
+if(USGSCSM_ENABLE_STARDS)
```
acpaquette: All the comments in this section need to be trimmed

CMakeLists.txt
```diff
@@ -718,31 +831,100 @@ target_include_directories(usgscsm
...
+                      ${PROJ_TARGET})
+
 add_executable(usgscsm_cam_test bin/usgscsm_cam_test.cc)
 target_link_libraries(usgscsm_cam_test
     usgscsm)
 
-install(TARGETS usgscsm
-        RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
-        LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}/csmplugins/
-        ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR})
```
acpaquette, line 745: This was changed to avoid linking errors with `usgscsm_cam_test`. Why did this get more or less reverted or fixed in some other way? It seems like if we don't it will introduce the issue again

bin/csm_translate.cc
```diff
@@ -0,0 +1,300 @@
...
+        std::cerr << "Unknown STARDS option '" << kv.first << "'\n";
+        return false;
+      }
+    }
+    writeUsgsCsmModelToStards(model, path, compression, blockSize, arrayThreshold);
+    return true;
+  }
+#else
+  (void)advanced;
+#endif
```
acpaquette, line 193: Why cast advanced to void?

bin/csm_translate.cc
```diff
@@ -0,0 +1,300 @@
...
+"  .stards         CSM model state in the STARDS binary format\n"
+#endif
+"\n"
+"Options:\n"
+"  -i <fmt>        Override the input format (json, msgpack, stards).\n"
+"  -o <fmt>        Override the output format (json, msgpack, stards).\n"
+"  --set key=val   Set an advanced, format-specific option (repeatable).\n"
+"                  STARDS output supports:\n"
+"                    compression=<none|gzip|zstd|lz4|gzip-shuffle|lz4-shuffle>\n"
+"                    block-size=<bytes>\n"
```
acpaquette, line 46: Not entirely sure why this would need to be set by a user. Might need a small explainer

src/Utilities.cpp
```diff
@@ -3053,3 +3109,257 @@ bool isUsgsCsmState(const std::string &str, std::string &modelName) {
...
+      }
+      case VariantMap::ValueType::VectorString: {
+        std::vector<std::string> v = vm.get<std::vector<std::string>>(key);
+        std::vector<size_t> shape{v.size()};
+        storeBySize(*ds, key, star::NDArray<std::string>(std::move(v), shape),
+                    arrayThreshold);
+        break;
+      }
+      default:
+        break;  // Unknown type: skip.
```
acpaquette: Might be nicer to have this output something. If this is called with some unaccounted for `VariantMap::ValueType` it would be good to know

tests/data/frameState.json.state
```diff
@@ -0,0 +1,158 @@
+{
```
acpaquette: Any particular reason to generate new data rather than using what was already in the data area? Could we generate the stards files from the existing test files?

acpaquette (commented):
13K+ lines is a bit much, did my best to find the things that seemed odd

CMakeLists.txt
Kelvinrr: I need to change it to make sure it only uses the submodule for WASM. afaik it'd be needed to at least resolve includes.

CMakeLists.txt
Kelvinrr: oh yeah I think I did that when testing

CMakeLists.txt
Kelvinrr: this was before I made that change I think, this PR was open for a while.

bin/csm_translate.cc
Kelvinrr: suppresses warnings since it's not used if STARDS is disabled.

bin/csm_translate.cc
Kelvinrr: Different compressions are useful in different contexts. Same with block size. the defaults are sensable but I think it's still worth being able to override. I mostly used it to test read/write times with different configurations to figure out the best defaults.

acpaquette: More specifically, we should add something to the help message to explain to what the blocksize does, compression is a bit more straight forward

CMakeLists.txt
Kelvinrr: make it on by default

src/UsgsAstroFrameSensorModel.cpp
```diff
@@ -15,8 +15,6 @@
 using json = nlohmann::json;
 
 // Declaration of static variables
-const std::string UsgsAstroFrameSensorModel::_SENSOR_MODEL_NAME =
-    "USGS_ASTRO_FRAME_SENSOR_MODEL";
```
acpaquette, line 19: Is there a particular reason these `_SENSOR_MODEL_NAME` definitions got moved to the .h files?

src/Utilities.cpp
```diff
@@ -2786,13 +2846,53 @@ VariantMap variantMapFromJson(const nlohmann::json& j) {
         result.set<std::vector<int>>(key, it->get<std::vector<int>>());
       } else if ((*it)[0].is_number()) {
         result.set<std::vector<double>>(key, it->get<std::vector<double>>());
+      } else if ((*it)[0].is_string()) {
+        result.set<std::vector<std::string>>(key, it->get<std::vector<std::string>>());
       }
+      // Arrays of objects or of arrays are dropped: a VariantMap cannot hold
+      // them, and stuffing in their serialized text would come back out of
+      // jsonFromVariantMap() as strings rather than the structure it was.
```
acpaquette: This seems a bit concerning that there is no warning or error to the user. I'm a little surprised we don't store the positions and velocities as a vector of vectors

tests/wasm/test_wasm.js
```diff
@@ -1,130 +0,0 @@
-/**
```
acpaquette, line 1: Was there a reason to remove these tests? Seems important to make sure the WASM module doesn't break

acpaquette (commented):
Overall looks good! A couple of small questions

src/UsgsAstroFrameSensorModel.cpp
Kelvinrr: It's been a while, but iirc it was causing symbol issues on windows.

src/Utilities.cpp
Kelvinrr: Smashing something to 1D is more memory efficient. StarDS does something similar to keep IOs as snappy as possible by minimizing memory fragmentation. I could add an exception here for object types.

tests/wasm/test_wasm.js
Kelvinrr: forgot to commit new tests.

src/Utilities.cpp
Kelvinrr: decided to add debug prints, I didn't want to muddy up the output.

acpaquette (approved):

PR #534, opened 2026-09-03
Do not truncate the reference time to whole seconds

oleg-alexandrov:
*getReferenceDateAndTime* cast the ephemeris time to *time_t* before formatting, in all four sensor models (frame, pushframe, linescan, SAR). That dropped the sub-second part, so images in the same second got an identical reference time, and thus an identical serial number when ISIS builds one from the *CsmInfo* group. jigsaw then rejected the framelets of a framing instrument such as JunoCam, spaced a few milliseconds apart, as duplicate serials. The fix keeps the time as a *double* so the sub-second part survives.

Verified by attaching a frame model to two JunoCam framelet cubes 0.003 s apart and running *csminit* then *getsn*: before, both serials ended in *.816Z*; after, they end in *.919Z* and *.922Z*, so they are distinct. The test suite was run locally.

This is the usgscsm side of the duplicate-serial problem in DOI-USGS/ISIS3#6134, which handles the ISIS side (preferring the instrument-based serial when the cube has an ISIS instrument label). It also covers a CSM camera with no ISIS instrument label, where the CSM serial is the only one available.

Disclaimer: done with AI assistance (Claude).

## Licensing


src/Utilities.cpp
Kelvinrr: does this need to be updated?

tests/FrameCameraTests.cpp
```diff
@@ -780,5 +780,18 @@ TEST_F(FrameSensorModel, losEllipsoidIntersect) {
...
+// JunoCam framelets are a few milliseconds apart, so their reference times (which
+// ISIS turns into serial numbers) must stay distinct. Bump the ephemeris time by
+// 3 ms and check the string changes; on the old whole-second code it would not.
+TEST_F(OrbitalFrameSensorModel, ReferenceDateTimeKeepsSubSecond) {
+  std::string dateA = sensorModel->getReferenceDateAndTime();
+  std::string state = sensorModel->getModelState();
+  size_t brace = state.find('{');
+  json j = json::parse(state.substr(brace));
+  j["m_ephemerisTime"] = j["m_ephemerisTime"].get<double>() + 0.003;
+  sensorModel->replaceModelState(state.substr(0, brace) + j.dump());
```
Kelvinrr: can we use the stateAsJson function here?

oleg-alexandrov: Done.

src/Utilities.cpp
oleg-alexandrov: Yeah, fixed.

Kelvinrr (approved):

